PR #8, opened 2020-07-14
Display diagnostics when wildcard should not exist in property key

xorye:
Fixes https://github.com/redhat-developer/quarkus-ls/issues/134

image

Signed-off-by: David Kwon <dakwon@redhat.com>

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
@@ -79,9 +79,7 @@ class MicroProfileValidator {
 	private void validateProperty(Property property) {
 		String propertyNameWithProfile = property.getPropertyNameWithProfile();
 		if (propertyNameWithProfile != null && !propertyNameWithProfile.isEmpty()) {
-			// Validate Syntax property
-			validateSyntaxProperty(propertyNameWithProfile, property);
-			// Validate Duplicate property
+			validateNoEqualsAfterProperty(propertyNameWithProfile, property);
 			validateDuplicateProperty(propertyNameWithProfile, property);
 		}
 
@@ -89,16 +87,15 @@ class MicroProfileValidator {
 		if (propertyName != null && !propertyName.isEmpty()) {
 			ItemMetadata metadata = MicroProfilePropertiesUtils.getProperty(propertyName, projectInfo);
 			if (metadata == null) {
-				// Validate Unknown property
 				validateUnknownProperty(propertyNameWithProfile, property);
 			} else {
-				// Validate property Value
 				validatePropertyValue(propertyNameWithProfile, metadata, property);
+				validateWildCardProperty(propertyNameWithProfile, metadata, property);
 			}
 		}
 	}
 
-	private void validateSyntaxProperty(String propertyName, Property property) {
+	private void validateNoEqualsAfterProperty(String propertyName, Property property) {
 		DiagnosticSeverity severity = validationSettings.getSyntax().getDiagnosticSeverity(propertyName);
 		if (severity == null) {
 			// The syntax validation must be ignored for this property name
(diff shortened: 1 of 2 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -161,6 +162,29 @@ private void validatePropertyValue(String propertyName, ItemMetadata metadata, P
...
+	 * 3. The corresponding property key in application.properties contains a wildcard 
+	 * 
+	 * In this case the corresponding property key should not have a wildcard
+	 * 
+	 * @param propertyKey the property key as it appears in application.properties
+	 * @param metadata    the corresponding property metadata
+	 * @param property    the property node
+	 */
+	private void validateWildCardProperty(String propertyKey, ItemMetadata metadata, Property property) {
+		if (!StringUtils.hasText(metadata.getExtensionName()) || !metadata.getName().contains("*")) {
```
xorye, line 179: Here, I use `!StringUtils.hasText(metadata.getExtensionName()` to determine if the current property metadata is a user-defined property (like: `@ConfigProperty(name = "greeting.hello{*}")`) or not.

Checking `!StringUtils.hasText(metadata.getExtensionName()` is not the best way to do it, maybe there is another way? Maybe we can have some sort of flag in `ItemMetadata` that is set when the jdt ls extension scans properties?

xorye:
In the Quarkus jdtls extension: https://github.com/redhat-developer/quarkus-ls/blob/e117c8fef04af3b785a0db580be991b1b97cac6d/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/providers/QuarkusConfigRootProvider.java#L532

and in MP jdtls extension: https://github.com/eclipse/lsp4mp/blob/731ff26413f7b536a8d7c672f320fbcb7d8c62fe/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/static-properties/mp-openapi-metadata.json#L64

we use the `{*}` syntax to represent mapped property.

Therefore for properties like: 
```
quarkus.log.category.{*}.handlers=test
mp.openapi.schema.{*}=test
```

this PR adds error ranges indicating that `{*}` should be replaced.

Also, in a Quarkus application, you can declare a config property with a `{*}` therefore I think we should not provide error ranges in this case.
```
@Path("/greeting")
public class GreetingResource {

    @ConfigProperty(name = "greeting.hello{*}")
    String message;
...
```

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
xorye: I have it as `contains("*")`  instead of `contains("{*}")` for now, because I think it would be interesting to cover this case:

```
kubernetes.annotations[*].key= test
```

xorye:
Updated:

image

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
xorye: In my latest push, I am checking against `metadata.getSource()` instead.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -80,7 +80,7 @@ private void validateProperty(Property property) {
 		String propertyNameWithProfile = property.getPropertyNameWithProfile();
 		if (propertyNameWithProfile != null && !propertyNameWithProfile.isEmpty()) {
 			// Validate Syntax property
```
datho7561, line 82: Remove this comment; the function name describes what the function does better.

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesDiagnosticsTest.java
```diff
@@ -746,4 +746,33 @@ public void validateMultilineKey() {
 				d(0, 0, 2, 4, "Unknown property 'qu.application.name'",
 						DiagnosticSeverity.Warning, ValidationType.unknown));
 	}
```
datho7561, line 748: Maybe add another test with
```java
quarkus.log.category.*.level=DEBUG
```
to show that this is also highlighted and to catch any regressions that change this.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
xorye: thank you, fixed

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesDiagnosticsTest.java
xorye: fixed

angelozerr:
> Therefore I think this PR is not necessary

I agree, thanks for your investigation @xorye .

PR #13, opened 2020-07-15
Model and completion for property expressions

datho7561:
In many MicroProfile implementations, `${other.property}` can be used to refer to the value of an already defined property when defining the value of another property.

This PR modifies the property file model to allow representing these property expressions. It also adds completion for defined properties when completion is opened after typing `${` as a part of a property value.

This PR was migrated from [quarkus-ls](https://github.com/redhat-developer/quarkus-ls/pull/340).


datho7561:
The tests appear to be working, but I identified this regression:
DollarCompletionRegression

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertyExpressionCompletionTest.java
```diff
@@ -0,0 +1,207 @@
...
+		testCompletionFor(text, true, c("${test.property}", "${test.property}", r(1, 22, 1, 23)));
+	}
+
+	@Test
+	public void beforeDollarSign() throws BadLocationException { // FIXME:
+		String text = //
+				"test.property = hello\n" + //
+				"other.test.property = |$\n";
+		testCompletionFor(text, true, c("${test.property}", "${test.property}", r(1, 22, 1, 23)));
+	}
```
datho7561, line 46: Should I look into changing this behaviour, or does this behaviour make sense?

xorye:
I think it would be great to add `$` to this list so that completion options are immediately proposed when `$` is typed: https://github.com/eclipse/lsp4mp/blob/97fa6bb483a9611c23a3b9f8a81a775623becbc6/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/settings/capabilities/ServerCapabilitiesConstants.java#L48

xorye:
I get a ClassCastException when I try the completion in an application.properties file that contains a comment:
```
#test
quarkus.application.name=name
quarkus.application.version=${|}
```
(I tried completion on the `|`)

Error:
```
Jul. 17, 2020 3:26:46 P.M. org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.ClassCastException: class org.eclipse.lsp4mp.model.Comments cannot be cast to class org.eclipse.lsp4mp.model.Property (org.eclipse.lsp4mp.model.Comments and org.eclipse.lsp4mp.model.Property are in unnamed module of loader 'app')
java.util.concurrent.CompletionException: java.lang.ClassCastException: class org.eclipse.lsp4mp.model.Comments cannot be cast to class org.eclipse.lsp4mp.model.Property (org.eclipse.lsp4mp.model.Comments and org.eclipse.lsp4mp.model.Property are in unnamed module of loader 'app')
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
	at java.base/java.util.concurrent.CompletableFuture.biApply(CompletableFuture.java:1238)
	at java.base/java.util.concurrent.CompletableFuture$BiApply.tryFire(CompletableFuture.java:1205)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:479)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
Caused by: java.lang.ClassCastException: class org.eclipse.lsp4mp.model.Comments cannot be cast to class org.eclipse.lsp4mp.model.Property (org.eclipse.lsp4mp.model.Comments and org.eclipse.lsp4mp.model.Property are in unnamed module of loader 'app')
	at org.eclipse.lsp4mp.services.MicroProfileCompletions.collectPropertyValueExpressionSuggestions(MicroProfileCompletions.java:412)
	at org.eclipse.lsp4mp.services.MicroProfileCompletions.doComplete(MicroProfileCompletions.java:113)
	at org.eclipse.lsp4mp.services.MicroProfileLanguageService.doComplete(MicroProfileLanguageService.java:84)
	at org.eclipse.lsp4mp.ls.ApplicationPropertiesTextDocumentService.lambda$completion$1(ApplicationPropertiesTextDocumentService.java:162)
	at java.base/java.util.concurrent.CompletableFuture.biApply(CompletableFuture.java:1236)
	... 7 more
```

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertyExpressionCompletionTest.java
xorye: I think in this case there shouldn't be completion options, right @fbricon, @angelozerr ?

datho7561:
> I think it would be great to add `$` to this list so that completion options are immediately proposed when `$` is typed:

Added

datho7561:
> I get a ClassCastException when I try the completion in an application.properties file that contains a comment:

Fixed; I added this as a unit test

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertyExpressionCompletionTest.java
datho7561: I addressed this in 01a207e

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyGraph.java
```diff
@@ -0,0 +1,92 @@
...
+	 * Gets a list of properties that do not depend on <code>property</code>.
+	 * 
+	 * These are the properties that, if a dependency of <code>property</code> on
+	 * them were introduced, wouldn't introduce a circular dependency.
+	 * 
+	 * @param property The property the find the independent properties of.
+	 * @return A list of all the properties whose value do not depend on this
+	 *         property.
+	 */
+	public List<String> independentProperties(String property) {
```
xorye, line 53: I wasn't sure if this method belongs here because I was thinking `PropertyGraph` is a simple Graph class that manages edges and could be reused for different applications (but maybe it's not very likely that we need graphs again).

But on second thought, maybe I'm wrong, since `PropertyGraph` was created specifically to determine relationships between properties and to solve problems like this "independent properties" problem.

datho7561: Hmm. I don't know either way as well. It would be easy enough to move `independentProperties` to a utilities class if it makes sense to do this.

datho7561:
Latest commit now suggests properties that appear in the properties file but haven't been defined elsewhere.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyGraph.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Queue;
+
+public class PropertyGraph {
```
datho7561, line 21: Add Javadoc

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyValueExpression.java
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	public String getReferencedPropertyName() {
+		Matcher m = PROPERTY_NAME_EXTRACTOR.matcher(getValue());
+		if (!m.matches()) {
+			return null;
+		}
+		return m.group(1);
+	}
+
+	public boolean isClosed() {
```
datho7561, line 60: Javadoc

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/parser/PropertiesParser.java
```diff
@@ -139,7 +145,7 @@ private void readComment() {
 	}
 
 	private enum StopReading {
-		PropertyName, PropertyValue
+		PropertyName, PropertyValue, PropertyExpression, PropertyValueLiteral
```
datho7561, line 148: Probably should be `PropertyValueExpression` for consistency

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/parser/PropertiesParser.java
```diff
@@ -307,27 +327,75 @@ private void readDelimiter() {
...
+				}
+			}
+		}
 		if (last != '\\') {
 			handler.endPropertyValue(this);
 			return true;
 		}
 		return false;
 	}
+	private boolean readPropertyExpression() {
```
datho7561, line 367: `readPropertyValueExpression()`

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/parser/PropertiesParser.java
```diff
@@ -307,27 +327,75 @@ private void readDelimiter() {
...
+		handler.startPropertyValueExpression(this);
+		read();
+		if (current != '{') {
+			handler.endPropertyValueExpression(this);
+			return true;
+		}
+		return continueReadPropertyExpression();
+	}
+
+	private boolean continueReadPropertyExpression() {
```
datho7561, line 377: `continueReadPropertyValueExpression`

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileCompletions.java
```diff
@@ -95,11 +97,29 @@ public CompletionList doComplete(PropertiesModel document, Position position, Mi
...
+				collectPropertyValueExpressionSuggestions(node, document, projectInfo, valuesRulesManager, completionSettings, list);
+			}
+			break;
+
 		case ASSIGN:
+			// Only collect if on right side of =
+			if (offset == node.getEnd()) {
+				collectPropertyValueSuggestions(node, document, projectInfo, valuesRulesManager, completionSettings, list);
+			}
+			break;
```
datho7561, line 115: This is a fix for https://github.com/eclipse/lsp4mp/issues/14; should I move it to a separate PR?

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileCompletions.java
```diff
@@ -412,6 +487,24 @@ private static CompletionItem getValueCompletionItem(ValueHint item, List<Conver
 		return completionItem;
 	}
 
+	/**
+	 * Make a completion item for a property given its metadata
+	 * 
+	 * @param property the metadata of the property to create a completion item for
+	 */
+	private static CompletionItem getPropertyCompletionItem(String propertyName, PropertyValueExpression propertyValueExpression, PropertiesModel model) {
+		String completionText = new StringBuilder("${").append(propertyName).append("}").toString();
```
datho7561, line 496: Just using `+` should be just as fast and easier to read.

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertyExpressionCompletionTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+	// Tests for where property expression completion should open
+
+	@Test
+	public void justDollarSignNoNewline() throws BadLocationException {
+		String text = //
+				"test.property = hello\n" + //
+						"other.test.property = $|";
+		testCompletionFor(text, generateInfoFor("test.property", "other.test.property"),
+				c("${test.property}", "${test.property}", r(1, 22, 1, 23)));
```
xorye, line 35: I think it's worth defining a new method in `MicroProfileAssert`:

```
	public static CompletionItem c(String text Range range) {
		return c(text, text, range);
	}
```
And using this new method instead of doing `c("${test.property}", "${test.property}", r(1, 22, 1, 23)));`

xorye: Instead of `r(1, 22, 1, 23)`, can we change it to `r(1, 22, 23)` throughout the PR?

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileCompletions.java
xorye: If we have tests for this case, I think it makes the most sense to move the fix and the tests to a separate PR, since the scope of this PR is already quite big.

datho7561: I took out the code from this PR and will make a separate PR for it.

xorye:
It seems like `@ConfigItem` properties like `quarkus.kafka-streams.application-id` can have property expressions in their default values (`${quarkus.application.name}`) and have it resolved properly.

But for config properties defined with `@ConfigProperty`, if the default value contains a property expression, the property expression doesn't get resolved. 

ie, if I have:
```
    @ConfigProperty(name = "greeting.message", defaultValue = "${greet}")
    String message;
```

and in application.properties:
```
greet=hi
```

I end up with `greeting.message=${greet}` instead of `greeting.message=hi`.

My question is whether or not we should take into account the dependencies for projectInfo properties. If we do, it would be ideal if we only consider the `ConfigItem` properties.

datho7561:
> It seems like `@ConfigItem` properties like `quarkus.kafka-streams.application-id` can have property expressions in their default values (`${quarkus.application.name}`) and have it resolved properly.

Oops. This PR doesn't take this into account. I assumed that the property expressions could only be used within the properties file. The dependencies are only built on properties referenced in the file. So if you did completion here:

```
quarkus.application.name = ${|
```

in a project that requires that Kafka configuration value, I would expect `quarkus.kafka-streams.application-id` as one of the completion items, which if selected for completion, would form a cyclic dependency.

It might be worthwhile to check this, since the cyclic dependency would invalidate the config file.

datho7561:
I did some testing, and it looks like Helidon doesn't resolve default values for property expressions:

```config.properties
# Application properties. This is the default greeting
app.greeting=Hello from ${combination.property},

# Microprofile server properties
server.port=8080
server.host=0.0.0.0

combination.property = ${server.host}:${server.port} ~ ${mp.openapi.scan.disable}
```

Output when accessing the resource, where `app.greeting` is a part of the message:
```
"Hello from 0.0.0.0:8080 ~ ${mp.openapi.scan.disable}, World!"
```

I'll test with Quarkus next to see what happens.

datho7561:
Doesn't work in Quarkus either:

```config.properties
# this is my configuration for quarkus
greeting.message=hello with ${quarkus.debug.reflection}, ${greeting.ve#getable[0].salad}
greeting.name=${greeting.message}
greeting.suffix=EXCLAMATION_MARK

greeting.ve#getable[0].salad = 42
```

When trying to launch the application with `./mvnw quarkus:dev` it errors, despite `quarkus.debug.reflection` begin `false` by default:

```
Caused by: java.util.NoSuchElementException: Property quarkus.debug.reflection not found
```

xorye:
In Quarkus, default values seem to be resolved for runtime properties like `quarkus.application.port`
image and not buildtime properties

xorye:
After discussing with @datho7561, maybe this PR should follow the property expressions defined here: https://github.com/eclipse/microprofile-config/pull/577 instead of the Quarkus property expressions defined here https://quarkus.io/guides/config#using-property-expressions?

https://github.com/eclipse/microprofile-config/pull/577 defines property expressions in the MP spec, but it's not merged yet.

If we decide to follow the proposed Microprofile spec, I think we should handle these cases:
image
in a different PR, in case this PR is getting too big/complicated.

@fbricon @angelozerr how does this sound?

angelozerr:
@datho7561 please fix conflicts

datho7561:
rebased

angelozerr:
When I circular dependency  with a like:

```
a=${a}
b=$|
```

When I try to open completion on b value with $, I have the following error:

```
[Error - 09:00:12] Request textDocument/completion failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: Cannot add self-loop edge on node a, as self-loops are not allowed. To construct a graph that allows self-loops, call allowsSelfLoops(true) on the Builder.
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:314)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:319)
	at java.base/java.util.concurrent.CompletableFuture.biApply(CompletableFuture.java:1238)
	at java.base/java.util.concurrent.CompletableFuture$BiApply.tryFire(CompletableFuture.java:1205)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:479)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:177)
Caused by: java.lang.IllegalArgumentException: Cannot add self-loop edge on node a, as self-loops are not allowed. To construct a graph that allows self-loops, call allowsSelfLoops(true) on the Builder.
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:217)
	at com.google.common.graph.ConfigurableMutableValueGraph.putEdgeValue(ConfigurableMutableValueGraph.java:82)
	at com.google.common.graph.ConfigurableMutableGraph.putEdge(ConfigurableMutableGraph.java:51)
	at org.eclipse.lsp4mp.model.PropertyGraph.<init>(PropertyGraph.java:67)
	at org.eclipse.lsp4mp.services.MicroProfileCompletions.collectPropertyValueExpressionSuggestions(MicroProfileCompletions.java:403)
	at org.eclipse.lsp4mp.services.MicroProfileCompletions.doComplete(MicroProfileCompletions.java:108)
	at org.eclipse.lsp4mp.services.MicroProfileLanguageService.doComplete(MicroProfileLanguageService.java:90)
	at org.eclipse.lsp4mp.ls.ApplicationPropertiesTextDocumentService.lambda$completion$1(ApplicationPropertiesTextDocumentService.java:166)
	at java.base/java.util.concurrent.CompletableFuture.biApply(CompletableFuture.java:1236)
	... 7 more
```

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyGraph.java
```diff
@@ -0,0 +1,130 @@
...
+	 * Gets a list of properties that do not depend on <code>property</code>.
+	 *
+	 * These are the properties that, if a dependency of <code>property</code> on
+	 * them were introduced, wouldn't introduce a circular dependency.
+	 *
+	 * @param property The property the find the independent properties of.
+	 * @return A list of all the properties whose value do not depend on this
+	 *         property.
+	 */
+	public List<String> independentProperties(String property) {
```
angelozerr, line 99: Rename to `getIndependentProperties`

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyValueExpression.java
```diff
@@ -0,0 +1,74 @@
...
+	 * Get the name of the referenced property, or null if brackets were missing in
+	 * the property expression.
+	 *
+	 * Does not check if the referenced property exists.
+	 *
+	 * @return the name of the referenced property, or null if brackets were missing
+	 *         in the property expression.
+	 */
+	public String getReferencedPropertyName() {
+		Matcher m = PROPERTY_NAME_EXTRACTOR.matcher(getValue());
```
angelozerr, line 57: Can we avoid using Regexp here just to extract property name? Using Regexp a lot, can grow up the memory or provide bad performance. @fbricon are you OK with that?

datho7561:
> When I have circular dependency with a like:
> 
> ```
> a=${a}
> b=$|
> ```

Do you think that it would be a good idea to suggest `a` as a completion option in this case?

angelozerr:
> Do you think that it would be a good idea to suggest a as a completion option in this case?

I think so. It's not the part of this PR, but we should have an error for a which higlight value ${a} to day that there is a circular problem.

datho7561:
I opened #61 for this case

angelozerr:
Congrats @datho7561! It's a very impressive PR.

PR #17, opened 2020-07-21
Hardcoded support for boolean converter

datho7561:
Boolean validator now supports 'Y', 'YES', '1', 'ON'. The error message for an invalid boolean value now displays a warning that the value will be automatically interpreted as 'false'.

Closes https://github.com/redhat-developer/quarkus-ls/issues/298

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
@@ -208,19 +208,32 @@ class MicroProfileValidator {
 			}
 		}
 
-		if ((metadata.isIntegerType() && !isIntegerString(value)) || (metadata.isFloatType() && !isFloatString(value))
-				|| (metadata.isBooleanType() && !isBooleanString(value))
+		if (metadata.isBooleanType() && !isBooleanString(value)) {
+			return "Type mismatch: " + metadata.getType() + " expected. By default, this value will be interpreted as 'false'";
+		}
+
+		if ((metadata.isIntegerType() && !isIntegerString(value)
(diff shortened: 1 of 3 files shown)


datho7561:
Screenshot from 2020-07-21 10-09-25

As mentioned above, I modified the error message to reflect the fact that, by default, unrecognised boolean values are interpreted as `false` by microprofile. Let me know if this change doesn't make sense to add.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -218,7 +221,7 @@ private static String getErrorIfValueTypeMismatch(ItemMetadata metadata, String
 	}
 
 	private static boolean isBooleanString(String str) {
-		return "true".equals(str) || "false".equals(str);
+		return "true".equals(str) || "false".equals(str) || "Y".equals(str) || "YES".equals(str) || "1".equals(str) || "ON".equals(str);
```
xorye, line 224: Can we have the check be case insensitive? If the value is `oN`, `on`, `TrUE`, `yes` etc., according to https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc#built-in-converters we should still consider those to be `true`

xorye:
Works well on my machine, with Quarkus

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
datho7561: Thanks! Forgot to look into this.

fbricon:
@datho7561 please fix the conflict

datho7561:
rebased

PR #19, opened 2020-07-22
Property expression definition

datho7561:
You can now trigger Go To Definition inside ${...}, and it will navigate to the location where the property value is assigned in the properties file.

Closes redhat-developer/quarkus-ls#333.

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
@@ -13,16 +13,16 @@
 *******************************************************************************/
 package org.eclipse.lsp4mp.services;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.LocationLink;
 import org.eclipse.lsp4j.Position;
-import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4mp.commons.MicroProfileProjectInfo;
 import org.eclipse.lsp4mp.commons.MicroProfilePropertyDefinitionParams;
 import org.eclipse.lsp4mp.commons.metadata.ItemHint;
@@ -36,6 +36,7 @@ import org.eclipse.lsp4mp.model.PropertiesModel;
 import org.eclipse.lsp4mp.model.Property;
 import org.eclipse.lsp4mp.model.PropertyKey;
 import org.eclipse.lsp4mp.model.PropertyValue;
+import org.eclipse.lsp4mp.model.PropertyValueExpression;
 import org.eclipse.lsp4mp.utils.MicroProfilePropertiesUtils;
 import org.eclipse.lsp4mp.utils.PositionUtils;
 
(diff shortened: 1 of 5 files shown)


xorye:
In this case, can we have the goto definition go to `city` instead of `%prod.city`?
goto_def_profile

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -108,6 +114,27 @@
 		return CompletableFuture.completedFuture(getEmptyDefinition(definitionLinkSupport));
 	}
 
+	private static CompletableFuture<Either<List<? extends Location>, List<? extends LocationLink>>>
```
datho7561, line 117: Javadoc

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -173,6 +200,24 @@ private static MicroProfilePropertyDefinitionParams getPropertyDefinitionParams(
...
+			Range propertyReferenceRange = new Range(document.positionAt(propertyValueExpression.getReferenceStartOffset()), document.positionAt(propertyValueExpression.getReferenceEndOffset()));
+			if (definitionLinkSupport) {
+				return Either.forRight(Collections.singletonList(
+					new LocationLink(document.getDocumentURI(), referencedPropertyRange, referencedPropertyRange, propertyReferenceRange)));
+			} else {
+				return Either.forLeft(Collections.singletonList(
+					new Location(document.getDocumentURI(), referencedPropertyRange)));
+			}
+		} catch (BadLocationException e) {
+			return null; //yum!!!
```
datho7561, line 217: Remove comment

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -108,6 +114,27 @@
 		return CompletableFuture.completedFuture(getEmptyDefinition(definitionLinkSupport));
 	}
 
+	private static CompletableFuture<Either<List<? extends Location>, List<? extends LocationLink>>>
+			findPropertyValueExpressionDefinition(PropertiesModel document, Node node, boolean definitionLinkSupport) {
```
datho7561, line 118: Change type of `node` to PropertyValueExpression

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -108,6 +114,27 @@
 		return CompletableFuture.completedFuture(getEmptyDefinition(definitionLinkSupport));
 	}
 
+	private static CompletableFuture<Either<List<? extends Location>, List<? extends LocationLink>>>
+			findPropertyValueExpressionDefinition(PropertiesModel document, Node node, boolean definitionLinkSupport) {
+		if (node.getNodeType() != NodeType.PROPERTY_VALUE_EXPRESSION) {
+			return CompletableFuture.completedFuture(getEmptyDefinition(definitionLinkSupport));
+		}
```
datho7561, line 121: Remove; this method only needs to work for `PropertyValueExpression`

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -173,6 +200,24 @@ private static MicroProfilePropertyDefinitionParams getPropertyDefinitionParams(
 		return definitionParams;
 	}
 
+	private static Either<List<? extends Location>, List<? extends LocationLink>>
+			getPropertyDefinition(PropertiesModel document, Property referencedProperty, PropertyValueExpression propertyValueExpression, boolean definitionLinkSupport) {
+		try {
+			Node key = referencedProperty.getKey();
+			Range referencedPropertyRange = new Range(document.positionAt(key.getStart()), document.positionAt(key.getEnd()));
```
datho7561, line 207: There is a helper function that can do this; instead of passing `document`, use the helper function.

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesDefinitionTest.java
```diff
@@ -26,6 +26,8 @@
  */
 public class ApplicationPropertiesDefinitionTest {
 
+	private static final String PROPERTY_DOCUMENT_NAME = "/microprofile.properties";
```
datho7561, line 29: This might not work on Windows

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
datho7561: After looking at it, its still more practical to also pass `document`.

angelozerr:
> In this case, can we have the goto definition go to city instead of %prod.city?

For this use case we should have 3 definitions I think.

datho7561:
> For this use case we should have 3 definitions I think (city, %prod.city and %dev.city), no?

I think this could be helpful, depending on how the client supports multiple definitions being sent from the server.

datho7561:
This is the behaviour in vscode if the server returns multiple definitions:

GoToAllDefinitions

I think being able to direct the user to the different definitions is helpful.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -36,6 +38,8 @@
 import org.eclipse.lsp4mp.model.Property;
 import org.eclipse.lsp4mp.model.PropertyKey;
 import org.eclipse.lsp4mp.model.PropertyValue;
+import org.eclipse.lsp4mp.model.PropertyValueExpression;
+import org.eclipse.lsp4mp.model.Node.NodeType;
```
angelozerr, line 42: Remove this unused import.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
```diff
@@ -180,20 +231,48 @@ private static MicroProfilePropertyDefinitionParams getPropertyDefinitionParams(
 		return definitionParams;
 	}
 
+	private static Either<List<? extends Location>, List<? extends LocationLink>> getPropertyDefinition(
+			PropertiesModel document, boolean definitionLinkSupport, PropertyValueExpression propertyValueExpression,
+			List<Property> referencedProperties) {
+
+		if (definitionLinkSupport) {
```
angelozerr, line 238: To avoid having 2 code for managing Location and LocationLink, it should be better to manage only LocationLink. The ApplicationTextDocumentService consumes the MicroProfileDefinition which returns LocationLink only and convert to Location if !definitionLinkSupport. 

We does that in LemMinX, see https://github.com/eclipse/lemminx/blob/d2d5bb76a1309f6d52bd8177d08aecbfafa5b5fb/org.eclipse.lemminx/src/main/java/org/eclipse/lemminx/XMLTextDocumentService.java#L335

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesDefinitionTest.java
angelozerr: Why? I'm in Windows OS and tests are working.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileDefinition.java
datho7561: Neither of these are unused any more.

PR #21, opened 2020-07-22
Property expression validation

datho7561:
Two things are validated:
 * If there is a closing brace
 * If the referenced property exists (if it is defined in a Java file or assigned a value in the properties file)

Closes redhat-developer/quarkus-ls#334

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
@@ -17,10 +17,13 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
 
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
@@ -33,6 +36,7 @@ import org.eclipse.lsp4mp.model.Node;
 import org.eclipse.lsp4mp.model.Node.NodeType;
 import org.eclipse.lsp4mp.model.PropertiesModel;
 import org.eclipse.lsp4mp.model.Property;
+import org.eclipse.lsp4mp.model.PropertyValueExpression;
 import org.eclipse.lsp4mp.model.values.ValuesRulesManager;
 import org.eclipse.lsp4mp.settings.MicroProfileValidationSettings;
 import org.eclipse.lsp4mp.utils.MicroProfilePropertiesUtils;
@@ -56,6 +60,7 @@ class MicroProfileValidator {
 
 	private final MicroProfileValidationSettings validationSettings;
 	private final Map<String, List<Property>> existingProperties;
+	private Set<String> allProperties;
 
 	public MicroProfileValidator(MicroProfileProjectInfo projectInfo, ValuesRulesManager valuesRulesManager,
 			List<Diagnostic> diagnostics, MicroProfileValidationSettings validationSettings) {
@@ -64,10 +69,12 @@ class MicroProfileValidator {
 		this.diagnostics = diagnostics;
 		this.validationSettings = validationSettings;
 		this.existingProperties = new HashMap<String, List<Property>>();
+		this.allProperties = null; // to be lazily init
 	}
 
 	public void validate(PropertiesModel document, CancelChecker cancelChecker) {
 		List<Node> nodes = document.getChildren();
+
 		for (Node node : nodes) {
 			if (cancelChecker != null) {
 				cancelChecker.checkCanceled();
@@ -100,6 +107,7 @@ class MicroProfileValidator {
 				// Validate property Value
 				validatePropertyValue(propertyNameWithProfile, metadata, property);
 			}
+			validatePropertyValueExpressions(property);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


datho7561:
WIP Demo:

PropertyExpressionValidation

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -161,6 +178,32 @@ private void validatePropertyValue(String propertyName, ItemMetadata metadata, P
...
+	private void validatePropertyValueExpressions(Property property, Set<String> allProperties) {
+		if (property.getValue() == null) {
+			return;
+		}
+		for (Node child : property.getValue().getChildren()) {
+			if (child != null && child.getNodeType() == NodeType.PROPERTY_VALUE_EXPRESSION) {
+				PropertyValueExpression propValExpr = (PropertyValueExpression) child;
+				String refdProp = propValExpr.getReferencedPropertyName();
+				if (!propValExpr.isClosed()) {
+					addDiagnostic("Missing '}'", propValExpr, DiagnosticSeverity.Error, ValidationType.missingCloseBrace.name());
```
datho7561, line 198: I wasn't sure whether it would be better to use `ValidationType.syntax` or to make a new `ValidationType`.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -63,20 +67,32 @@ public MicroProfileValidator(MicroProfileProjectInfo projectInfo, ValuesRulesMan
...
+		// Collect names of all properties defined in the configuration file and the project information
+		Set<String> allProperties = new HashSet<>();
+		allProperties.addAll(projectInfo.getProperties().stream().map((ItemMetadata info) -> {
+			return info.getName();
+		}).collect(Collectors.toList()));
+		allProperties.addAll(nodes.stream().filter(n -> {
+			return n.getNodeType() == NodeType.PROPERTY;
+		}).map(prop -> {
+			return ((Property)prop).getPropertyNameWithProfile();
+		}).collect(Collectors.toList()));
```
xorye, line 80: I think its best if we don't do this here. If we are in an application.properties file that has no property expressions, then it seems that we didn't need to compute this in the first place.

Maybe we can:

Check in `validatePropertyValueExpressions()` if the property is a project property by checking `MicroProfilePropertiesUtils.getProperty(propertyName, projectInfo) == null`
 like here:
https://github.com/eclipse/lsp4mp/blob/82c4798ac142efeb06499ece053407ef01009d46/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java#L90-L91

and if `MicroProfilePropertiesUtils.getProperty(propertyName, projectInfo) == null`, then we can check if it matches a property in the properties file.

What do you think?

datho7561: I think this is a good idea. I'll work on getting this implemented.

datho7561: I switched to using lazy loading for creating the list (only create the list right before its needed, then reuse it for future tasks). The implementation is kind of messy; I am open to suggestions on how to improve it.

fbricon:
@datho7561 please make sure new file headers follow this format:

> /********************************************************************************
>  * Copyright (c) {date} {owner}[ and others]
>  * 
>  * This program and the accompanying materials are made available under the
>  * terms of the Eclipse Public License v. 2.0 which is available at
>  * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
>  * which is available at https://www.apache.org/licenses/LICENSE-2.0.
>  * 
>  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
> 
> ********************************************************************************/

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/ValidationType.java
```diff
@@ -24,7 +24,7 @@
  */
 public enum ValidationType {
 
-	syntax, unknown, duplicate, value, required, requiredValue;
+	syntax, unknown, duplicate, value, required, requiredValue, missingCloseBrace, unknownReference;
```
angelozerr, line 27: The validation type is bound with settings `microprofile.tools.validation.*` https://github.com/redhat-developer/vscode-microprofile#supported-vs-code-settings. I think we should:

 * remove missingCloseBrace and replace it with syntax.
 * rename `unknownReference` with `expression` and after that we should expose a new settings in vscode-microprofile.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -167,8 +175,46 @@ private void validatePropertyValue(String propertyName, ItemMetadata metadata, P
...
+						return n.getNodeType() == NodeType.PROPERTY;
+					}).map(prop -> {
+						return ((Property)prop).getPropertyNameWithProfile();
+					}).collect(Collectors.toList()));
+				}
+				String refdProp = propValExpr.getReferencedPropertyName();
+				if (!propValExpr.isClosed()) {
+					addDiagnostic("Missing '}'", propValExpr, DiagnosticSeverity.Error, ValidationType.missingCloseBrace.name());
+				}
+				if (!allProperties.contains(refdProp)) {
```
angelozerr, line 208: Severity must be get with :

`DiagnosticSeverity severity = validationSettings.getSyntax().getDiagnosticSeverity(propertyName);`

You should test at first if severity is != none to do the validation (to avoid doing validation for nothing). See validateSyntaxProperty method for instance.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -167,8 +175,46 @@ private void validatePropertyValue(String propertyName, ItemMetadata metadata, P
...
+					allProperties.addAll(projectInfo.getProperties().stream().map((ItemMetadata info) -> {
+						return info.getName();
+					}).collect(Collectors.toList()));
+					allProperties.addAll(property.getOwnerModel().getChildren().stream().filter(n -> {
+						return n.getNodeType() == NodeType.PROPERTY;
+					}).map(prop -> {
+						return ((Property)prop).getPropertyNameWithProfile();
+					}).collect(Collectors.toList()));
+				}
+				String refdProp = propValExpr.getReferencedPropertyName();
```
angelozerr, line 204: Get severity `DiagnosticSeverity severity = validationSettings.getSyntax().getDiagnosticSeverity(propertyName);`

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
angelozerr: Use ValidationType.syntax (see my comments about settings).

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/ValidationType.java
datho7561: Opened https://github.com/redhat-developer/vscode-microprofile/pull/18 for the client changes for the new setting.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileValidator.java
```diff
@@ -167,8 +175,53 @@ private void validatePropertyValue(String propertyName, ItemMetadata metadata, P
...
+						}).collect(Collectors.toList()));
+						allProperties.addAll(property.getOwnerModel().getChildren().stream().filter(n -> {
+							return n.getNodeType() == NodeType.PROPERTY;
+						}).map(prop -> {
+							return ((Property)prop).getPropertyNameWithProfile();
+						}).collect(Collectors.toList()));
+					}
+					String refdProp = propValExpr.getReferencedPropertyName();
+					if (!allProperties.contains(refdProp)) {
+						addDiagnostic("Unknown referenced property '" + refdProp + "'", propValExpr, expressionSeverity, ValidationType.expression.name());
```
angelozerr, line 212: For the range I think we should higlight the error of the expression without ${ and } like higlight does.

datho7561: What about for the missing `}`?

angelozerr: Keep like today

PR #4, opened 2020-07-13
Rename settings prefix to microprofile

rzgry:
Changes prefix of settings and client commands to `microprofile`

See https://github.com/redhat-developer/quarkus-ls/issues/325

Signed-off-by: Ryan Zegray <ryan.zegray@ibm.com>

PR #22, opened 2020-07-23
No completion suggestions on left side of `=`

### LINKED ISSUE #14: Completion in properties file gives enum values before `=`
Given the following simple application:

```java
package org.acme;

import java.util.Optional;

import javax.ws.rs.GET;
import javax.ws.rs.Path;
import javax.ws.rs.Produces;
import javax.ws.rs.core.MediaType;

import org.eclipse.microprofile.config.inject.ConfigProperty;

@Path("/greeting")
public class GreetingResource {

    @ConfigProperty(name = "greeting.message")
    String message;

    @ConfigProperty(name = "greeting.suffix")
    Suffix suffix;

    @ConfigProperty(name = "greeting.name")
    Optional<String> name;

    @GET
    @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
        return message + " " + name.orElse("world") + suffix;
    }

}
```

Along with this defined `enum`:

```java
package org.acme;

public enum Suffix {
    EXCLAMATION_MARK, PERIOD, QUESTION_MARK;
    @Override
    public String toString() {
        switch (this) {
        case EXCLAMATION_MARK:
            return "!";
        case PERIOD:
            return ".";
        case QUESTION_MARK:
            return "?";
        default:
            return "";
        }
    }
}
```

When completion is opened in the microprofile property file between the property name and `=` (where the `|` is):

```
greeting.message=hello
greeting.suffix|=
```

The values for the enum (`EXCLAMATION_MARK,` `PERIOD`, `QUESTION_MARK`) are suggested.

When one the values is selected, it is added between the property name and the `=`, making the property file invalid.

```
greeting.message=hello
greeting.suffixEXCLAMATION_MARK=
```

Enum values should either not be suggested before `=`, or the range where the enum gets inserted should be adjusted to after the `=`.

datho7561:
Fixes #14

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileCompletions.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileCompletions.java
@@ -98,6 +98,11 @@ class MicroProfileCompletions {
 			// no completions
 			break;
 		case ASSIGN:
+			// Only collect if on right side of =
+			if (offset >= node.getEnd()) {
+				collectPropertyValueSuggestions(node, document, projectInfo, valuesRulesManager, completionSettings, list);
+			}
+			break;
 		case PROPERTY_VALUE:
 			// completion on property value
 			collectPropertyValueSuggestions(node, document, projectInfo, valuesRulesManager, completionSettings, list);
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesCompletionTest.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesCompletionTest.java
@@ -24,7 +24,7 @@ import org.junit.Test;
 
 /**
  * Test with completion in 'application.properties' file.
- * 
+ *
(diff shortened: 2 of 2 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/MicroProfileCompletions.java
```diff
@@ -96,6 +96,11 @@ public CompletionList doComplete(PropertiesModel document, Position position, Mi
 			// no completions
 			break;
 		case ASSIGN:
+			// Only collect if on right side of =
+			if (offset == node.getEnd()) {
```
xorye, line 100: I think we probably want `offset >= node.getEnd()` because if the user has `microprofile.tools.formatting.surroundEqualsWithSpaces` on, they will probably trigger completion after a whitespace character `quarkus.http.cors = |`

image

MP settings don't work on master at the moment though, we need https://github.com/eclipse/lsp4mp/pull/4

datho7561: Good catch. I added `>=`, and added this test case.

fbricon:
@datho7561 please fix the conflict

angelozerr:
@datho7561 please fix the conflict

datho7561:
rebased

PR #47, opened 2020-08-10
Fix missing unit in Gauge metrics snippet

rzgry:
Adds unit to the `@Gauge` snippet since it is required https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/app-programming-model.adoc#gauge

All others provide a default value so it is not needed. https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/app-programming-model.adoc#fields

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/resources/org/eclipse/lsp4mp/snippets/mp-metrics.json
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/resources/org/eclipse/lsp4mp/snippets/mp-metrics.json
@@ -37,7 +37,8 @@
     "body": [
       "@Gauge(",
       "\tname = \"${1:name}\",",
-      "\tdescription = \"${2:description}\"",
+      "\tdescription = \"${2:description}\",",
+      "\tunit = ${3|MetricUnits.NONE,MetricUnits.BYTES,MetricUnits.SECONDS,MetricUnits.PERCENT,\"\"|}",
       ")"
     ],
     "description": "Denotes a gauge, which samples the value of the annotated object.",


fbricon:
"Fix missing comma"?

rzgry:
This PR adds the `unit` attribute in the `@Guage` annotation snippet since it was missing: `"\tunit = $3",`. The added comma in the first line is needed just to add the new attribute.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/resources/org/eclipse/lsp4mp/snippets/mp-metrics.json
```diff
@@ -37,7 +37,8 @@
     "body": [
       "@Gauge(",
       "\tname = \"${1:name}\",",
-      "\tdescription = \"${2:description}\"",
+      "\tdescription = \"${2:description}\",",
+      "\tunit = $3",
```
angelozerr, line 41: unit is a string, so I think it should be 

`"\tunit = \"${3:unit}\"",`

no?

rzgry: I think for the majority of use cases the MetricUnit constants are used. https://github.com/eclipse/microprofile-metrics/blob/master/api/src/main/java/org/eclipse/microprofile/metrics/MetricUnits.java#L34

Ex. from https://quarkus.io/guides/microprofile-metrics
`@Gauge(name = "highestPrimeNumberSoFar", unit = MetricUnits.NONE, description = "Highest prime number so far.")`

So i didn't want to wrap the placeholder in quotes if they would need to remove them to use the MetricUnit constants. 

I would think we leave it as is, or maybe change it to something like: `"\tunit = ${3:\"unit\"}",`?

angelozerr: Snippets should generate code without syntax error, that's why I wonder why it misses quote, but I understand your point of view. So I suggest that you use snippet choice https://microsoft.github.io/language-server-protocol/specification#choice which provides MetricUnits.NONE, and other value of MetricUnits and double quote.

rzgry: Updated to use the choice syntax. I only added a couple options. Not sure if we should cover all possible MetricUnits 
options? Seems like it might be too much.

image

angelozerr: I think it's nice now because snippets generates annotation without syntax error. Thanks @rzgry !

PR #52, opened 2020-08-12
Fix duplicate of `quarkus-properties` when registering `textDocument/rangeFormatting`

angelozerr:
Fix duplicate of `quarkus-properties` when registering `textDocument/rangeFormatting`

Fixes https://github.com/redhat-developer/vscode-quarkus/issues/296

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/settings/capabilities/MicroProfileCapabilityManager.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/settings/capabilities/MicroProfileCapabilityManager.java
@@ -54,7 +54,6 @@ public class MicroProfileCapabilityManager {
 	private final LanguageClient languageClient;
 
 	private ClientCapabilitiesWrapper clientWrapper;
-	private TextDocumentRegistrationOptions formattingRegistrationOptions;
 
 	private final List<IMicroProfileRegistrationConfiguration> registrationConfigurations;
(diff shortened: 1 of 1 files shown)


fbricon:
Can you explain what happened, and change the commit message (it's a fix, no longer an investigation)

angelozerr:
The problem is that formatting and rangeFormatting used the same instance of List https://github.com/eclipse/lsp4mp/blob/7b84e7f6d0685a83a870966cdfdccb52cb6b5802/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/settings/capabilities/MicroProfileCapabilityManager.java#L113 

as Registration options https://github.com/eclipse/lsp4mp/blob/7b84e7f6d0685a83a870966cdfdccb52cb6b5802/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/settings/capabilities/MicroProfileCapabilityManager.java#L144

When configure was done for this shared list in https://github.com/eclipse/lsp4mp/blob/7b84e7f6d0685a83a870966cdfdccb52cb6b5802/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/settings/capabilities/MicroProfileCapabilityManager.java#L147 the first call of `registerCapability` from formatting added `quarkus-properties` and the second call of `registerCapability` from rangeFormatting added a second `quarkus-properties` in the shared list.

To fix the issue, formatting and rangeFormatting must not share the same List of  options.

fbricon:
ok makes sense

fbricon:
@datho7561 please review this one

PR #92, opened 2020-10-02
Add sequence diagrams

### LINKED ISSUE #75: Sequence diagram of properties completion
Discussed in #71 

Would be nice to have a sequence diagram of the communication between lsp4mp, the client and the external component in our architecture documentation.

Also would be nice to have a simple diagram with vscode completion which shows MP completion and LSP4MP server side which returns Spring Boot Metadata JSON list of properties.

rzgry:
Fixes #75 

Adds sequence diagrams that I have been working on for a talk for eclipsecon. Would love to get any feedback on if anything should be changed.

Signed-off-by: Ryan Zegray <ryan.zegray@ibm.com>


docs/architecture.md
```diff
@@ -54,6 +54,11 @@ For example:
 
 6. The MicroProfile Language Server receives the project information, adds it to its cache, and returns the completion options stored in the project information as the response to the original `textDocument/completion` request.
 
+In VS Code, the MicroProfile JDT is used as the external component, so the following diagram represents the communication sequence between components
+
+![Completion cache hit sequence](./images/property-completion-cache-hit-sequence.png)
```
angelozerr, line 59: I love your diagram 👍 

But I think we should change the order of the images (to follow the real process because the first time cache is empty)

rzgry: Thanks! Re-ordered them to put the cache miss first.

PR #104, opened 2020-11-06
Replace String.isBlank by String.trim().isEmpty() for Java8 compatibi…

### LINKED ISSUE #103: Server does not run under Java8
While testing with IntelliJ, I noticed some errors due to unknown methods.
I also tried to rebuild under Java8 and got some compilation errors.

jeffmaury:
…lity

Fixes #103

Signed-off-by: Jeff MAURY <jmaury@redhat.com>


fbricon:
You need to change the java requirement to Java 8 in the quarkus ls pom.xml, so we can detect breakages

angelozerr:
It should be nice to add for instance in https://github.com/eclipse/lsp4mp/tree/master/microprofile.ls and https://github.com/eclipse/lsp4mp/blob/master/CONTRIBUTING.md that code must be written in Java 8 for MicroProfile LS

jeffmaury:
The pom already has 1.8 requirement but it does seem to relate only to syntax not APIs !!!!

fbricon:
That's what the release flag is for.

PR #106, opened 2020-11-13
Use propertyName when checking unknown property exclusion list.

rgrunber:
- A valid property exclusion pattern will fail to match against
  propertyNameWithProfile due to the profile prefix not matching
- Resolves redhat-developer/vscode-quarkus#314


microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/ApplicationPropertiesDiagnosticsTest.java
```diff
@@ -230,6 +230,15 @@ public void validateUnknownPropertiesExcludedWithPattern() throws BadLocationExc
 		// * pattern --> all errors are ignored
 		unknown.setExcluded(new String[] { "*" });
 		testDiagnosticsFor(value, 0, getDefaultMicroProfileProjectInfo(), settings);
+
+		value = "kafka-streams.cache.max.bytes.buffering=10240\n" + //
```
angelozerr, line 234: As your test doesn't use the original value, I suggest that you create a new test method like `validateUnknownPropertiesWithProfileExcludedWithPattern`

rgrunber: Done.

angelozerr:
Thanks @rgrunber !

PR #115, opened 2020-12-02
Replace 'ps -p' with 'kill -0' in ParentProcessWatcher

### LINKED ISSUE #110: Use `kill -0` instead of `ps -p` in ParentProcessWatcher
To prevent some crashes when running in WSL Alpine Linux (see https://github.com/redhat-developer/vscode-java/issues/1711), we need to use `kill -0` instead of `ps -p` in [ParentProcessWatcher](https://github.com/eclipse/lsp4mp/blob/d13833cb6b95fb3955425a7b08b989089024c269/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/commons/ParentProcessWatcher.java#L91) like in https://github.com/eclipse/eclipse.jdt.ls/pull/1612/files

fbricon:
Fixes #110


rgrunber:
I've tested this out and seems to be working well, whether the UI is closed (one of parents is terminated) or a SIGKILL is sent.

PR #116, opened 2020-12-04
Separate Java/Properties features in MicroProfile LS

### LINKED ISSUE #89: Separate Java/Properties features in MicroProfile LS
MicroProfile LS manages Java and properties files. For LSP TextDocumentService layer, it's pretty well separated, we have

 * ApplicationPropertiesTextDocumentService.java for properties files. We should be renamed to PropertiesTextDocumentService (because it manages all properties files like microprofile-config.properties, aplication.properties and application.yaml).

 * JavaTextDocumentService.java for java files
The LSP TextDocumentService should called a language service. We have that for ApplicationPropertiesTextDocumentService (it consumes MicroProfileLanguageService but not for java).

The idea of this issue is :

move MicroProfileLanguageService to the org.eclipse.lsp4mp.services.properties package. Rename MicroProfileLanguageService to MicroProfilePropertiesLanguageService

create MicroProfileJavaLanguageService which will be consumed by JavaTextDocumentService.java

angelozerr:
Separate Java/Properties features in MicroProfile LS

Fixes #89

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
@@ -26,6 +26,7 @@
             <command id="microprofile/java/fileInfo"/>
             <command id="microprofile/java/codeAction"/>
             <command id="microprofile/java/codeLens"/>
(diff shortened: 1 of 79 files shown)


angelozerr:
Bad PR

PR #123, opened 2020-12-10
Convert MavenProjectName from enum to static class.

rgrunber:
- Related redhat-developer/vscode-quarkus#319
- Rename enum MavenProjectName to static class MicroProfileProjectName
  so that it may be extended in other places
- Replace references of MavenProjectName with a String


microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/BasePropertiesManagerTest.java
```diff
@@ -58,42 +58,33 @@
...
+		public static String microprofile_health_quickstart="microprofile-health-quickstart";
+		public static String microprofile_lra="microprofile-lra";
+		public static String microprofile_context_propagation="microprofile-context-propagation";
+		public static String microprofile_metrics="microprofile-metrics";
+		public static String microprofile_opentracing="microprofile-opentracing";
+		public static String microprofile_openapi="microprofile-openapi";
+		public static String microprofile_reactive_messaging="microprofile-reactive-messaging";
+		public static String microprofile_graphql="microprofile-graphql";
+		public static String microprofile_jwt_quickstart="microprofile-jwt-quickstart";
+		public static String quarkus_container_images="quarkus-container-images";
```
rzgry, line 84: I have an issue open to move test project names that are only used in quarkus ls over to the quarkus-ls repo #7

Would be nice to remove these from lsp4mp and include them as part of quarkus-ls.

ex. quarkus-container-images: https://github.com/redhat-developer/quarkus-ls/blob/7f2f111b725b9590dc89373293f375cb033ccd5f/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus.test/src/main/java/com/redhat/microprofile/jdt/quarkus/containerimages/QuarkusContainerImagesTest.java#L35

rgrunber: I moved `kubernetes` and `quarkus_container_images` as those were the ones I found whose references were only on the quarkus-ls side. If I've missed any, let me know. I also removed the corresponding projects in the lsp4mp tests since presumably they were using the project on the quarkus-ls side anyways. The tests seem to be passing.

I've also modified GradleProjectName in the same way but I haven't created a quarkus equivalent because I couldn't find anything that is used only on the quarkus-ls side.

angelozerr:
Thanks @rgrunber for your great job!

Thanks @rzgry for your comments.

PR #124, opened 2020-12-11
Support arbitrary number of member values in PropertiesHoverParticipant.

rgrunber:
We should add support for any amount of member values for a given
annotation in addition to continued support for a single member value
and an optional default value member.

- Related redhat-developer/quarkus-ls#376

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
@@ -16,7 +16,9 @@ package org.eclipse.lsp4mp.jdt.core.java.hover;
 import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.getAnnotation;
 import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.getAnnotationMemberValue;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -53,9 +55,9 @@ public class PropertiesHoverParticipant implements IJavaHoverParticipant {
 
 	private final String annotationName;
 
-	private final String annotationMemberName;
-
 	private final String defaultValueAnnotationMemberName;
+
+	private final String [] annotationMembers;
 
 	public PropertiesHoverParticipant(String annotationName, String annotationMemberName) {
 		this(annotationName, annotationMemberName, null);
@@ -73,8 +75,24 @@ public class PropertiesHoverParticipant implements IJavaHoverParticipant {
 	 */
 	public PropertiesHoverParticipant(String annotationName, String annotationMemberName,
 			String defaultValueAnnotationMemberName) {
+		this(annotationName, new String [] { annotationMemberName }, defaultValueAnnotationMemberName);
+	}
+
+	/**
+	 * The definition participant constructor.
+	 *
+	 * @param annotationName                   the annotation name (ex :
+	 *                                         io.quarkus.scheduler.Scheduled)
+	 * @param annotationMembers                a list of annotation members (ex :
+	 *                                         cron, every, etc.
+	 * @param defaultValueAnnotationMemberName the annotation member name for
+	 *                                         default value and null otherwise.
+	 *
+	 */
+	public PropertiesHoverParticipant(String annotationName, String [] annotationMembers,
+			String defaultValueAnnotationMemberName) {
 		this.annotationName = annotationName;
-		this.annotationMemberName = annotationMemberName;
+		this.annotationMembers = annotationMembers;
 		this.defaultValueAnnotationMemberName = defaultValueAnnotationMemberName;
 	}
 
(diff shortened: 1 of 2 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
```diff
@@ -74,8 +76,13 @@ public PropertiesHoverParticipant(String annotationName, String annotationMember
 	public PropertiesHoverParticipant(String annotationName, String annotationMemberName,
 			String defaultValueAnnotationMemberName) {
 		this.annotationName = annotationName;
-		this.annotationMemberName = annotationMemberName;
 		this.defaultValueAnnotationMemberName = defaultValueAnnotationMemberName;
+		this.annotationMembers = new String [] { annotationMemberName };
+	}
+
+	public PropertiesHoverParticipant(String annotationName, String [] annotationMembers) {
```
rgrunber, line 83: I should add a Javadoc here to explain the new 'annotationMembers' parameter.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
```diff
@@ -126,6 +140,9 @@ public Hover collectHover(JavaHoverContext context, IProgressMonitor monitor) th
 			return null;
 		}
 
+		// property references may be surrounded by curly braces in Java file
```
rgrunber, line 143: Is it legal for property names to contain curly braces ? If not, this should be pretty safe to do in all cases. Otherwise we can polish this up.

rgrunber (commented):
The idea is to pass in via the new constructor a list of supported member values (eg. new String [] { "cron", "every", "delay", "delayUnit" } and get property hover support.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
rgrunber: In the case of `@ConfigProperty`, you could (if you wanted to) have something like `name = {greeting.message}` where in application.properties `{greeting.message}=hello` . In fact it doesn't work exactly since hovering over the property in the Java file says the property is not set, but in the application file there is no validation error for it.

So I think we should be fine treating curly braces as special in `@Scheduled` member values since we really do need them as the value can be a literal or a property reference.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
rgrunber: Done.

angelozerr:
Is there any chance to have some tests with your feature please?

rgrunber:
> Is there any chance to have some tests with your feature please?

Sure, but I think it would have to be done in quarku-ls, much like MicroProfileConfigJavaHoverTest. The purpose of the API is to simply be used on the quarkus side. I'll post the corresponding contribution (which will include the new QuarkusScheduledHoverParticipant) with a testcase.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
```diff
@@ -74,8 +76,21 @@ public PropertiesHoverParticipant(String annotationName, String annotationMember
...
 		this.defaultValueAnnotationMemberName = defaultValueAnnotationMemberName;
+		this.annotationMembers = new String [] { annotationMemberName };
+	}
+
+	/**
+	 * The definition participant constructor.
+	 *
+	 * @param annotationName                   the annotation name (ex :
+	 *                                         io.quarkus.scheduler.Scheduled)
+	 * @param annotationMembers                a list of annoation members (ex :
```
angelozerr, line 88: annotation

angelozerr: @rgrunber what about having a main constructor like:

```java
PropertiesHoverParticipant(String annotationName, String [] annotationMemberNames, String [] annotationMemberValues)
```

and after you provide another constructor which call this main constructor?

rgrunber: Done. the initialization is done by a single constructor.

rgrunber:
Seems we have a failure in :
```
java.lang.NoSuchMethodError: 'void org.eclipse.m2e.jdt.internal.BuildPathManager.scheduleDownload(org.eclipse.jdt.core.IPackageFragmentRoot, org.eclipse.m2e.core.embedder.ArtifactKey, boolean, boolean)'
	at org.eclipse.lsp4mp.jdt.core.faulttolerance.MicroProfileFaultTolerancePropertiesTest.microprofileFaultTolerancePropertiesTest(MicroProfileFaultTolerancePropertiesTest.java:42)
```
Do we need to bump jdt.ls.version ?

angelozerr:
> I'll check to see if a simple update to that property would fix this.

Thanks!

rgrunber:
test this please

angelozerr:
Thanks a lot @rgrunber !

PR #125, opened 2021-01-12
Move MicroProfileForJavaAssert into internal.core class.

rgrunber:
- Related #376
- Move MicroProfileForJavaAssert from the restricted
  org.eclipse.lsp4mp.jdt.internal.core.java into the already exported
  org.eclipse.lsp4mp.jdt.internal.core .

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/META-INF/MANIFEST.MF
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/META-INF/MANIFEST.MF
@@ -32,5 +32,4 @@ Export-Package: org.eclipse.lsp4mp.jdt.core,
  org.eclipse.lsp4mp.jdt.core.metrics,
  org.eclipse.lsp4mp.jdt.core.openapi,
  org.eclipse.lsp4mp.jdt.core.opentracing,
- org.eclipse.lsp4mp.jdt.core.restclient,
- org.eclipse.lsp4mp.jdt.internal.core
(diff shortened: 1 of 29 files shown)


angelozerr:
@rgrunber I think it should be better to move those assert class (MicroProfileAssert and MicroProfileForJavaAssert) in `org.eclipse.lsp4mp.jdt.core` and remove export package of `org.eclipse.lsp4mp.jdt.internal.core` and adapt Quarkus test with this package rename.

rgrunber:
> @rgrunber I think it should be better to move those assert class (MicroProfileAssert and MicroProfileForJavaAssert) in `org.eclipse.lsp4mp.jdt.core` and remove export package of `org.eclipse.lsp4mp.jdt.internal.core` and adapt Quarkus test with this package rename.

Only `MicroProfileAssert` isn't needed but I've moved it as well. This will require adding a few import package changes on the quarkus-ls side as this was used there via the internal.core package.

`org.eclipse.lsp4mp.jdt.internal.core.java` should be gone now, but org.eclipse.lsp4mp.jdt.internal.core` will still exist as there are 4 other files in internal.core.

angelozerr:
@rgrunber please remove the export package of  `org.eclipse.lsp4mp.jdt.internal.core` (andafter that you will see the  benefit to move MicroProfileAssert in the core).

rgrunber:
JavaUtils is now moved as well. This was enough to eliminate internal.core from exports.

angelozerr:
Thanks @rgrunber !

PR #126, opened 2021-01-19
Improve range finding logic for annotation member values.

rgrunber:
- Related redhat-developer/quarkus-ls#376
- @Foo (bar = {val}, baz = {val}) would fail when hovering over the
  2nd {val} because the index-of matching would detect the first
  {val}. Instead, we match against the target member (baz) and begin
  our index-of matching from the end of that term.

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
@@ -19,6 +19,8 @@ import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.getAnnotationMem
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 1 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
```diff
@@ -130,13 +132,17 @@ public Hover collectHover(JavaHoverContext context, IProgressMonitor monitor) th
...
 			if (propertyKey != null) {
 				ISourceRange r = ((ISourceReference) annotation).getSourceRange();
-				int offset = annotationSource.indexOf(propertyKey);
-				propertyKeyRange = utils.toRange(typeRoot, r.getOffset() + offset, propertyKey.length());
-
-				if (!hoverPosition.equals(propertyKeyRange.getEnd())
-						&& Ranges.containsPosition(propertyKeyRange, hoverPosition)) {
-					found = true;
-					break;
+				Pattern memberPattern = Pattern.compile(".*[^\\w{\\.](" + annotationMemberName + ")[^\\w}\\.].*", Pattern.DOTALL);
```
angelozerr, line 135: If you have this

`@Sheduled(cron="cron")`, the pattern returns the second cron (inside the string).

I would exclude " instead of {:

`Pattern memberPattern = Pattern.compile(".*[^\\w"\\.](" + annotationMemberName + ")[^\\w"\\.].*", Pattern.DOTALL);`

rgrunber: I've added `\"` to the regular expression to catch this.

rgrunber:
Updated with your suggested regular expression pattern.

angelozerr:
Thanks @rgrunber !

PR #134, opened 2021-03-29
Remove pack200 support.

rgrunber:
- https://eclip.se/572043

See https://wiki.eclipse.org/Tycho/Pack200 for why tycho-p2-plugin can also be removed.


rgrunber:
Mickael thinks the p2-metadata part might be needed so let's hold off on merging this for now.

rgrunber:
I think this should be fine. The docs clearly state p2-metadata was added in this case to detect the newly created packed artifacts. Also the documentation I'm seeing for using eclipse-jarsigner-plugin doesn't indicate the need for it.

PR #138, opened 2021-04-01
Release OSGi bundles with qualifiers.

rgrunber:
- Since snapshot releases (including qualifiers) are available, we
  should preserve these in the release to ensure a proper upgrade path.


rgrunber:
@fbricon , @datho7561 . Thoughts ? I think this should allow us to preserve qualifiers on the OSGi side directly from release. We can probably trigger a 0.2.1 build afterwards.

fbricon:
I think it might work

fbricon (approved):

PR #144, opened 2021-06-30
Recognize `application.yml` as a configuration source

datho7561:
Most features for YAML configuration files already work in both `application.yml` and `application.yaml`. However, this PR fixes the following features which were previously only working for `application.yaml`:

 * Hovering a `@ConfigProperty` annotation's `name` field shows the value that's defined in the config file, along with a link to the config file.
 * Invoking "Go to definition" on a `@ConfigProperty` annotation's `name` field opens the configuration file

Closes redhat-developer/quarkus-ls#373

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProject.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProject.java
@@ -38,12 +38,15 @@ public class JDTMicroProfileProject {
 	public static final String APPLICATION_PROPERTIES_FILE = "application.properties";
 	@Deprecated
 	public static final String APPLICATION_YAML_FILE = "application.yaml";
+	@Deprecated
+	public static final String APPLICATION_YML_FILE = "application.yml";
 
 	private final List<IConfigSource> configSources;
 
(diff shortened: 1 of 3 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProject.java
```diff
@@ -38,12 +38,15 @@
 	public static final String APPLICATION_PROPERTIES_FILE = "application.properties";
 	@Deprecated
 	public static final String APPLICATION_YAML_FILE = "application.yaml";
+	@Deprecated
```
angelozerr, line 41: Why you add @Deprecated ?

datho7561: I think the idea is that, eventually, we should move the code for handling `application.properties` and `application.yaml` to quarkus-ls, since these files are Quarkus specific.

angelozerr:
@datho7561 I thinkthis PR is replaced with https://github.com/eclipse/lsp4mp/pull/149 ?

datho7561:
@angelozerr I can incorporate the changes in this PR into that PR.

datho7561:
All of the changes in the PR have been applied to https://github.com/eclipse/lsp4mp/pull/149

PR #149, opened 2021-07-13
Extension point for contributing configuration sources

datho7561:
lsp4mp extensions can now contribute implementations of IConfigSourceProvider, which is a class that generates a list of IConfigSource given a project. All the IConfigSource generated by all the IConfigSourceProvider will be used for language features that use a property's value (eg. hover).

This means that if an implementation of MicroProfile provides a particular configuration file that is not a part of the base MicroProfile config, an extension to lsp4mp can be implemented which introduces support for the configuration file, without having to modify lsp4mp directly.

As example of how this can be used, this opens the door to moving the implementation and inclusion of the Quarkus `application.properties` and `application.yml` configuration sources out of lsp4mp, and instead providing them through a seperate project (quarkus-ls).

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
@@ -11,6 +11,9 @@
    <extension-point id="projectLabelProviders" 
                     name="%projectLabelProviders.name" 
                     schema="schema/projectLabelProviders.exsd"/>
+   <extension-point id="configSourceProviders"
+                    name="%configSourceProviders.name"
+                    schema="schema/configSourceProviders.exsd"/>
 
    <!-- Delegate command handler for application.properties file -->
    <extension point="org.eclipse.jdt.ls.core.delegateCommandHandler">
@@ -45,6 +48,12 @@
       <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.MavenProjectLabelProvider" />
       <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.GradleProjectLabelProvider" />
       <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.MicroProfileProjectLabelProvider" />
+   </extension>
+
+   <!-- Configuration sources -->
+   <extension point="org.eclipse.lsp4mp.jdt.core.configSourceProviders">
+      <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.DefaultMicroProfilePropertiesConfigSourceProvider" />
+      <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.QuarkusConfigSourceProvider" />
    </extension>
 
    <!-- Java Util Logging Config support -->
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/schema/configSourceProviders.exsd
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/schema/configSourceProviders.exsd
@@ -1,0 +1,104 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Schema file written by PDE -->
+<schema targetNamespace="org.eclipse.lsp4mp.jdt.core" xmlns="http://www.w3.org/2001/XMLSchema">
+   <annotation>
+      <appinfo>
+         <meta.schema plugin="org.eclipse.lsp4mp.jdt.core" id="configSourceProviders" name="%configSourceProviders.name"/>
+      </appinfo>
+      <documentation>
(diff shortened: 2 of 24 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/schema/configSourceProviders.exsd
```diff
@@ -0,0 +1,104 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="since"/>
+      </appinfo>
+      <documentation>
+         0.3.0
```
datho7561, line 72: 0.4.0

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
```diff
@@ -47,6 +50,12 @@
       <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.MicroProfileProjectLabelProvider" />
    </extension>
 
+   <!-- Configuration sources -->
+   <extension point="org.eclipse.lsp4mp.jdt.core.configSourceProviders">
+      <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.DefaultMicroProfilePropertiesConfigSourceProvider" />
+      <provider class="org.eclipse.lsp4mp.jdt.internal.core.providers.QuarkusConfigSourceProvider" />
```
angelozerr, line 56: QuarkusConfigSourceProvider should be hosted in quarkus-ls project but perhaps you want to do that in an another PR?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProject.java
```diff
@@ -119,4 +107,12 @@ public Integer getPropertyAsInteger(String key, Integer defaultValue) {
 				.collect(Collectors.toList());
 	}
 
+	private List<IConfigSource> getConfigSources() {
+		TreeSet<IConfigSource> configSourceSet = new TreeSet<>((a, b) -> b.getOrdinal() - a.getOrdinal());
```
angelozerr, line 111: I think it can be a good idea to cache the list because getConfigSources will be consumeda lot of time.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
datho7561: Yes, I was planning to do this in another PR, since moving the implementation means rewriting a lot of the test cases. Also, I need the SNAPSHOT of this build to be published in order to recognize the extension point in quarkus-ls

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProject.java
datho7561: The reason I am not caching the sources is because for https://github.com/redhat-developer/vscode-quarkus/issues/371 I need to reload the list of configuration sources when a new `application-*.properties` file is added/deleted. Do you have an idea for a strategy to cache these files?

datho7561, line 111: I can't use a Set here, since there can be entries with the same ordinal

datho7561: Maybe I could make a `static` config source provider and a `dynamic` config source provider, and only reload the dynamic config sources. Does this sound like a good idea?

angelozerr: For the moment we will skip the cache because it's fast.

angelozerr: >Do you have an idea for a strategy to cache these files?

We must implement implement IResourceChangeListener if I remember, but let's skip cache for the moment.

angelozerr:
It works great and code is very clean. Thanks @datho7561 !

PR #152, opened 2021-07-19
Upversion to 0.3.1

datho7561:


datho7561:
@rgrunber

rgrunber:
Ah, so the reason it's 2 changes smaller than usual is because the build auto-sets the tags now. Cool. Just running to verify.

datho7561:
The build always autotagged the release commit. It doesn't push the changes it makes to the main branch, though, just the detached commit of the released code.

PR #153, opened 2021-07-20
Fix pom file versions that build failed to commit.

rgrunber:
- Build correctly adjusted pom files but failed to include them
- globbing provided to git-add should be quoted to prevent this


datho7561 (approved):
LGTM

PR #155, opened 2021-07-22
Validate defaultValue of @ConfigProperty

### LINKED ISSUE #90: Validate defaultValue of @ConfigProperty
Validate defaultValue of @ConfigProperty  according the field type.

Given this sample:

```java
@ConfigProperty(name = "greeting.message" defaultValue="XXXXXXXXXXXXXXX")
Integer intValue;
```

We should have `XXXXXXXXXXXXXXX` which is highlighted as error because the expected value is an Integer.

This feature must be implemented on JDT LS side with the `com.redhat.microprofile.jdt.core.javaFeatureParticipants` diagnostics extension point.

As we have validation for value type in MicroProfile LS for application.properties, it should be good to move all possible validator features in the `org.eclipse.lsp4mp.microprofile.commons` package.

rgrunber:
- Use documentation from org.eclipse.microprofile.config.spi.Converter
- Fixes #90

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
@@ -1,0 +1,137 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.jdt.internal.config.java;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.Signature;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.ASTVisitor;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.Expression;
+import org.eclipse.jdt.core.dom.FieldDeclaration;
+import org.eclipse.jdt.core.dom.ITypeBinding;
+import org.eclipse.jdt.core.dom.NormalAnnotation;
+import org.eclipse.jdt.core.dom.StringLiteral;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.DiagnosticSeverity;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants;
+import org.eclipse.lsp4mp.jdt.core.java.diagnostics.IJavaDiagnosticsParticipant;
+import org.eclipse.lsp4mp.jdt.core.java.diagnostics.JavaDiagnosticsContext;
+import org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils;
+import org.eclipse.lsp4mp.jdt.internal.metrics.MicroProfileMetricsConstants;
+
+/**
+ * Validate "defaultValue" attribute of @ConfigProperty and generate
+ * diagnostics if "defaultValue" cannot be represented by the given
+ * field type.
+ *
+ * See https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc
+ * for more details on default converters.
+ */
+public class MicroProfileConfigPropertyDiagnosticsParticipant implements IJavaDiagnosticsParticipant {
+
+	@Override
+	public List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProgressMonitor monitor)
+			throws CoreException {
+		List<Diagnostic> diagnostics = new ArrayList<>();
+		validateClass(diagnostics, context, monitor);
+		return diagnostics;
+	}
+
+	private void validateClass(List<Diagnostic> diagnostics, JavaDiagnosticsContext context,
+			IProgressMonitor monitor) {
+		CompilationUnit ast = context.getASTRoot();
+		ast.accept(new ConfigPropertiesAnnotationValidator(diagnostics, context));
+	}
+
+	public class ConfigPropertiesAnnotationValidator extends ASTVisitor {
+		private List<Diagnostic> diagnostics;
+		private JavaDiagnosticsContext context;
+
+		public ConfigPropertiesAnnotationValidator(List<Diagnostic> diagnostics, JavaDiagnosticsContext context) {
+			this.diagnostics = diagnostics;
+			this.context = context;
+		}
+
+		@Override
+		public boolean visit(NormalAnnotation node) {
+			if (AnnotationUtils.isMatchAnnotation(node, MicroProfileConfigConstants.CONFIG_PROPERTY_ANNOTATION)) {
+				try {
+					Expression defValueExpr = AnnotationUtils.getAnnotationMemberValueExpression(node, MicroProfileConfigConstants.CONFIG_PROPERTY_ANNOTATION_DEFAULT_VALUE);
+					ASTNode parent = node.getParent();
+					if (defValueExpr != null && parent instanceof FieldDeclaration) {
+						String defValue = ((StringLiteral) defValueExpr).getLiteralValue();
+						FieldDeclaration field = (FieldDeclaration) parent;
+						ITypeBinding fieldBinding = field.getType().resolveBinding();
+						if (fieldBinding != null && !isAssignable(fieldBinding, defValue)) {
+							Range range = context.getUtils().toRange(context.getTypeRoot(), defValueExpr.getStartPosition(), defValueExpr.getLength());
+							diagnostics.add(new Diagnostic(range,
+									"'" + defValue + "'" + " does not match the expected type of '" + fieldBinding.getName() + "'.",
+									DiagnosticSeverity.Error, MicroProfileMetricsConstants.DIAGNOSTIC_SOURCE));
+						}
+					}
+				} catch (JavaModelException e) {
+					return false;
+				}
+			}
+			return false;
+		}
+
+		private boolean isAssignable(ITypeBinding fieldBinding, String defValue) {
+			String fqn = Signature.getTypeErasure(fieldBinding.getQualifiedName());
+			try {
+				switch (fqn) {
+				case "boolean":
+				case "java.lang.Boolean":
+					return Boolean.valueOf(defValue) != null;
+				case "byte":
+				case "java.lang.Byte":
+					return Byte.valueOf(defValue) != null;
+				case "short":
+				case "java.lang.Short":
+					return Short.valueOf(defValue) != null;
+				case "int":
+				case "java.lang.Integer":
+					return Integer.valueOf(defValue) != null;
+				case "long":
+				case "java.lang.Long":
+					return Long.valueOf(defValue) != null;
+				case "float":
+				case "java.lang.Float":
+					return Float.valueOf(defValue) != null;
+				case "double":
+				case "java.lang.Double":
+					return Double.valueOf(defValue) != null;
+				case "char":
+				case "java.lang.Character":
+					return Character.valueOf(defValue.charAt(0)) != null;
+				case "java.lang.Class":
+					return  Class.forName(defValue) != null;
+				case "java.lang.String":
+					return true;
+				default:
+					return false;
+				}
+			} catch (NumberFormatException | ClassNotFoundException e) {
+				return false;
+			}
+		}
+	}
+}
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/MicroProfileConfigJavaDiagnosticsTest.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/MicroProfileConfigJavaDiagnosticsTest.java
@@ -1,0 +1,59 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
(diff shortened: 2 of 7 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.commons;
```
angelozerr, line 14: This class should be hosted in https://github.com/eclipse/lsp4mp/tree/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+		private JavaDiagnosticsContext context;
+
+		public ConfigPropertiesAnnotationValidator(List<Diagnostic> diagnostics, JavaDiagnosticsContext context) {
+			this.diagnostics = diagnostics;
+			this.context = context;
+		}
+
+		@Override
+		public boolean visit(NormalAnnotation node) {
+			if (AnnotationUtils.isMatchAnnotation(node, "ConfigProperty")) {
```
angelozerr, line 67: Please use constant from https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/MicroProfileConfigConstants.java

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+				try {
+					Expression defValueExpr = AnnotationUtils.getAnnotationMemberValueExpression(node, "defaultValue");
```
angelozerr, line 69: Please use constant from https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/MicroProfileConfigConstants.java

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/MicroProfileConfigJavaDiagnosticsTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.eclipse.lsp4mp.jdt.core.utils.IJDTUtils;
+import org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants;
+import org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceErrorCode;
+import org.eclipse.lsp4mp.jdt.internal.metrics.MicroProfileMetricsConstants;
+import org.junit.Test;
+
+/**
+ * MicroProfile Fault Tolerance definition in Java file.
+ *
+ * @author Angelo ZERR
```
angelozerr, line 25: Update the author

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/MicroProfileConfigJavaDiagnosticsTest.java
```diff
@@ -0,0 +1,54 @@
+package org.eclipse.lsp4mp.jdt.core.config;
```
angelozerr, line 1: It misses the header EPL license.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+import org.eclipse.jdt.core.dom.StringLiteral;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4j.DiagnosticSeverity;
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants;
+import org.eclipse.lsp4mp.jdt.core.java.diagnostics.IJavaDiagnosticsParticipant;
+import org.eclipse.lsp4mp.jdt.core.java.diagnostics.JavaDiagnosticsContext;
+import org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils;
+import org.eclipse.lsp4mp.jdt.internal.metrics.MicroProfileMetricsConstants;
+
```
angelozerr, line 39: Add a comment which explains which validation this class does.

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/MicroProfileConfigJavaDiagnosticsTest.java
rgrunber: I've just kept it empty without author reference. Let me know if you prefer I list myself or anything like that. I've seen some files with / some without.

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/config/MicroProfileConfigJavaDiagnosticsTest.java
rgrunber: Thanks. Good catch.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,137 @@
...
+					Expression defValueExpr = AnnotationUtils.getAnnotationMemberValueExpression(node, MicroProfileConfigConstants.CONFIG_PROPERTY_ANNOTATION_DEFAULT_VALUE);
+					String defValue = ((StringLiteral) defValueExpr).getLiteralValue();
+					ASTNode parent = node.getParent();
+					if (parent instanceof FieldDeclaration) {
+						FieldDeclaration field = (FieldDeclaration) parent;
+						ITypeBinding fieldBinding = field.getType().resolveBinding();
+						if (fieldBinding != null && !isAssignable(fieldBinding, defValue)) {
+							Range range = context.getUtils().toRange(context.getTypeRoot(), defValueExpr.getStartPosition(), defValueExpr.getLength());
+							diagnostics.add(new Diagnostic(range,
+									"'" + defValue + "'" + " does not match the expected type of '" + fieldBinding.getName() + "'.",
```
rgrunber, line 86: As discussed, I should probably match this error message to the one used in the application properties file for the mismatched types.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,137 @@
...
+			this.diagnostics = diagnostics;
+			this.context = context;
+		}
+
+		@Override
+		public boolean visit(NormalAnnotation node) {
+			if (AnnotationUtils.isMatchAnnotation(node, MicroProfileConfigConstants.CONFIG_PROPERTY_ANNOTATION)) {
+				try {
+					Expression defValueExpr = AnnotationUtils.getAnnotationMemberValueExpression(node, MicroProfileConfigConstants.CONFIG_PROPERTY_ANNOTATION_DEFAULT_VALUE);
+					String defValue = ((StringLiteral) defValueExpr).getLiteralValue();
```
angelozerr, line 78: You should check if defValueExpr is !=null to avoid having NPE with the usecase

```java
@ConfigProperty(name = "greeting.message")
    String message;
```

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
angelozerr: Keep like this for the moment because I find your error message better than properties file:

. In application.properties : `Type mismatch: int expected`
* In Java : `'xx' does not match the expected type of 'int'`

I think it's better and application.properties should follow this message.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,137 @@
...
+					ASTNode parent = node.getParent();
+					if (parent instanceof FieldDeclaration) {
+						FieldDeclaration field = (FieldDeclaration) parent;
+						ITypeBinding fieldBinding = field.getType().resolveBinding();
+						if (fieldBinding != null && !isAssignable(fieldBinding, defValue)) {
+							Range range = context.getUtils().toRange(context.getTypeRoot(), defValueExpr.getStartPosition(), defValueExpr.getLength());
+							diagnostics.add(new Diagnostic(range,
+									"'" + defValue + "'" + " does not match the expected type of '" + fieldBinding.getName() + "'.",
+									DiagnosticSeverity.Error, MicroProfileMetricsConstants.DIAGNOSTIC_SOURCE));
```
angelozerr, line 87: Just a note,in properties file we can configure the severity 

image

In this PR we can keep like this, but I wonder if we should use the same severity than properties file or define a new settings for java file?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
rgrunber: Bah, I just changed it. Let me revert. It's an easy enough change.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
rgrunber: Done.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
angelozerr: ok no problem.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
```diff
@@ -0,0 +1,137 @@
...
+			return false;
+		}
+
+		private boolean isAssignable(ITypeBinding fieldBinding, String defValue) {
+			String fqn = Signature.getTypeErasure(fieldBinding.getQualifiedName());
+			try {
+				switch (fqn) {
+				case "boolean":
+				case "java.lang.Boolean":
+					return Boolean.valueOf(defValue) != null;
```
angelozerr, line 103: There is a problem with boolean value (you can set any value, and you will have not error).

It's because `Boolean.valueOf("XXXXX") `return `false` and not null. Please check boolean like https://github.com/eclipse/lsp4mp/blob/master/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java#L291

rgrunber: Ah, could it be that `PropertiesFileValidator.isBooleanString(..)` is too strict ? According to https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc :

`boolean and java.lang.Boolean , values for true (case insensitive) "true", "1", "YES", "Y" "ON". Any other value will be interpreted as false`

So even "XXXXX" is technically just false. If that's the case, then any String value should be supported.

angelozerr: I think we should follow the specification, so you are right and PropertiesFileValidator.isBooleanString should be less strict. Thanks @rgrunber for your clarification.

angelozerr:
Thanks so much @rgrunber. We have a nice validation feature and very good sample how to validate attribute of annotation.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java
rgrunber: I tihnk there's a good argument to be made for maybe changing this setting to "Validation severity for property values for MicroProfile." This way we can include things like "*.properties" and "*.java". However, does the microprofile.jdt extension code even have access to the shared settings from microprofile.ls ? I didn't see a way to access them.

PR #156, opened 2021-07-28
Do not include fallback annotated method as its own fallback option

### LINKED ISSUE #148: Exclude the method that's being annotated when showing completion for fallback method
eg.

```java
@GET
@Produces(MediaType.TEXT_PLAIN)
@Fallback(fallbackMethod="|")
public String faultTolerant() {
    return "hello, world!";
}
```

If you trigger completion at the `|`, you shouldn't see `faultTolerant` suggested.

rgrunber:
- Fixes #148
- Adjust tests accordingly


angelozerr:
Thanks @rgrunber !

PR #157, opened 2021-07-30
Update eclipse-jarsigner-plugin

datho7561:
The version of the jarsigner plugin that we are using tries to connect to build.eclipse.org, which is being phased out. This updated version should work with the new build infrastrucutre.


angelozerr:
Thanks @datho7561

PR #158, opened 2021-07-30
Warn when a @ConfigProperty is declared but no value is assigned

### LINKED ISSUE #146: @ConfigProperty validation when property is not set
Given this @ConfigProperty:

```java
public class InventoryConfig {

    @ConfigProperty(name="port")
    private int port;

}
```
In this case the port property is not set (we can see that with hover).

It should be nice to highlight as an error (warn) that port is not set. It will give the capability:

 * to have an overwiew of several properties which are not set when we open the Java file
 * bind code actions to configure it. See https://github.com/eclipse/lsp4mp/issues/147

datho7561:
Closes #146


angelozerr:
@datho7561 I think you should use https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigPropertyDiagnosticsParticipant.java (rename MicroProfileConfigPropertyDiagnosticsParticipant to MicroProfileConfigDiagnosticsParticipant) and add your own validation.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
```diff
@@ -72,13 +72,15 @@
...
    <extension point="org.eclipse.lsp4mp.jdt.core.javaFeatureParticipants">
-       <!-- Java definition for the MicroProfile @ConfigProperty annotation -->
+      <!-- Java definition for the MicroProfile @ConfigProperty annotation -->
       <definition class="org.eclipse.lsp4mp.jdt.internal.config.java.MicroProfileConfigDefinitionParticipant" />      
-       <!-- Java hover for the MicroProfile @ConfigProperty annotation -->
-      <hover class="org.eclipse.lsp4mp.jdt.internal.config.java.MicroProfileConfigHoverParticipant" />   
+      <!-- Java hover for the MicroProfile @ConfigProperty annotation -->
+      <hover class="org.eclipse.lsp4mp.jdt.internal.config.java.MicroProfileConfigHoverParticipant" />  
+      <!-- Java validation for the MicroProfile @ConfigProperty annotation -->
+      <diagnostics class="org.eclipse.lsp4mp.jdt.internal.config.java.MicroProfileConfigDiagnosticsParticipant" />
```
angelozerr, line 80: Please remove https://github.com/eclipse/lsp4mp/blob/466ab11e11a715f38c241d22df0681ab2d6a6f60/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml#L222

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/MicroProfileTextDocumentService.java
```diff
@@ -114,8 +114,11 @@ public void didOpen(DidOpenTextDocumentParams params) {
 	@Override
 	public void didChange(DidChangeTextDocumentParams params) {
 		TextDocumentService service = getTextDocumentService(params.getTextDocument());
-		if (service != null) {
+		if (service == javaTextDocumentService) {
```
datho7561, line 117: This is the wrong location to revalidate

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigDiagnosticsParticipant.java
```diff
@@ -0,0 +1,241 @@
...
+import org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants;
+import org.eclipse.lsp4mp.jdt.core.java.diagnostics.IJavaDiagnosticsParticipant;
+import org.eclipse.lsp4mp.jdt.core.java.diagnostics.JavaDiagnosticsContext;
+import org.eclipse.lsp4mp.jdt.core.project.JDTMicroProfileProject;
+import org.eclipse.lsp4mp.jdt.core.project.JDTMicroProfileProjectManager;
+import org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils;
+import org.eclipse.lsp4mp.jdt.core.utils.JDTTypeUtils;
+
+/**
+ * Collects diagnostics related to the <code>@ConfigProperty</code> annotation
```
angelozerr, line 53: Explain more which validation is supported (if I remember Roland explained that)

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigDiagnosticsParticipant.java
```diff
@@ -0,0 +1,268 @@
...
+		 * @param annotation       the ConfigProperty annotation
+		 * @param defaultValueExpr the default value expression, or null if no default
+		 *                         value is defined
+		 */
+		private void validatePropertyHasValue(Annotation annotation, Expression defaultValueExpr) {
+			try {
+				String name = null;
+				Expression nameExpression = null;
+
+				nameExpression = getAnnotationMemberValueExpression(annotation, CONFIG_PROPERTY_ANNOTATION_NAME);
```
angelozerr, line 179: `Expression nameExpression = getAnnotationMemberValueExpression(annotation, CONFIG_PROPERTY_ANNOTATION_NAME);`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigDiagnosticsParticipant.java
```diff
@@ -0,0 +1,268 @@
...
+			return false;
+		}
+	}
+
+	private static boolean doesPropertyHaveValue(String property, JavaDiagnosticsContext context) {
+		IJavaProject javaProject = context.getJavaProject();
+		try {
+			JDTMicroProfileProject mpProject = JDTMicroProfileProjectManager.getInstance()
+					.getJDTMicroProfileProject(javaProject);
+			return mpProject.getPropertyInformations(property).size() > 0;
```
angelozerr, line 261: I will add a new method `hasProperty`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
```diff
@@ -159,6 +163,12 @@ public boolean visit(IResourceDelta delta) throws CoreException {
 					event.setProjectURIs(new HashSet<String>());
 					event.getProjectURIs().add(JDTMicroProfileUtils.getProjectURI(file.getProject()));
 					fireAsyncEvent(event);
+				} else if (isConfigSource(file) && isFileContentChanged(delta)) {
+					MicroProfilePropertiesChangeEvent event = new MicroProfilePropertiesChangeEvent();
```
angelozerr, line 167: Why you need that? It's for refreshing validation in Java file when application.properties changed? I find it's really shame to do that because you will compute all properties from Java sources when application.properties is saved although we will have none changed for properties.

I wonder if we should manage that in MP LSside when didSave is called, we publish diagnostics for all opened java files.

datho7561: I tried the strategy you mentioned (calling on `didSave`) but since the info is pulled from the target application.properties, it wouldn't actually update properly unless you saved twice (I guess once to trigger JDT to put the application.properties in target, one more for microprofile-jdt to update). The same thing was true for running on `didChange`, since the target is only updated on save. If I run on `didSave` and `didChange` it wasn't too bad, but the diagnostics were still noticeably out of sync of the content of application.properties.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/utils/StringUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * This class is a copy/paste from
+ * https://github.com/spring-projects/spring-framework/blob/243f2890ee9e98c97c8dc7278f033def3bf33f86/spring-core/src/main/java/org/springframework/util/StringUtils.java
+ * which contains only method required by {@link AntPathMatcher}.
+ *
+ * @author Angelo ZERR
+ *
+ */
+public class StringUtils {
```
angelozerr, line 31: You should have org.eclipse.lsp4mp.commons.util.StringUtils  in MP LSside too https://github.com/eclipse/lsp4mp/tree/master/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/commons (commons package have the same classes for mp.jdt and mp.ls)

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/diagnostics/JavaDiagnosticsContext.java
```diff
@@ -31,15 +34,33 @@
...
 	}
 
+	/**
+	 * Returns the MicroProfileJavaDiagnosticsSettings.
+	 *
+	 * Should not be null.
+	 *
+	 * @return the MicroProfileJavaDiagnosticsSettings
+	 */
+	public MicroProfileJavaDiagnosticsSettings getSettings() {
```
angelozerr, line 60: Let's keep like this, but this settings should be store on JDT LS side as singleton.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigDiagnosticsParticipant.java
```diff
@@ -0,0 +1,266 @@
...
+				String name = null;
+				Expression nameExpression = getAnnotationMemberValueExpression(annotation, CONFIG_PROPERTY_ANNOTATION_NAME);
+				boolean hasDefaultValue = defaultValueExpr != null;
+
+				if (nameExpression != null) {
+					name = ((StringLiteral) nameExpression).getLiteralValue();
+				}
+
+				if (name != null && !hasDefaultValue && !isPropertyIgnored(name)
+						&& !doesPropertyHaveValue(name, context)) {
```
angelozerr, line 185: No needto check if property is ignored if there are none diagnostics In other words, I would move doesPropertyHaveValue before isPropertyIgnored

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
angelozerr: Ok that's maake sense, but perhaps we should manage an another type (ex : MicroProfilePropertiesScope.ONLY_PROPERTIES) to avoid evict sources properties on MP LS side an drecompute the sources properties. Let's keep like this and please create an issue for that.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
```diff
@@ -187,6 +197,29 @@ private boolean isJavaFile(IFile file) {
 			return JAVA_FILE_EXTENSION.equals(file.getFileExtension());
 		}
 
+		private boolean isConfigSource(IFile file) {
+			IPath path = file.getProjectRelativePath();
+			// we want the config update on the class path not the sources
+			if (path.segmentCount() <= 2
```
angelozerr, line 203: It's annoying to hard code the output folder. I will remove this code and I will loop for each IConfigSourceProvider and I will add a new method `IConfigSource#isSameFile(IPath path)` and isSameFile should use https://github.com/eclipse/lsp4mp/blob/f9afc25b0ba3083251fd60cbca976bb3603d0750/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/AbstractConfigSource.java#L70

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/diagnostics/JavaDiagnosticsContext.java
datho7561: We talked about making a SharedSettings for the jdt-ls side. Is this what you mean?

angelozerr: Yes exactly

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
datho7561: https://github.com/eclipse/lsp4mp/issues/162

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/diagnostics/JavaDiagnosticsContext.java
datho7561: https://github.com/eclipse/lsp4mp/issues/163

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/AbstractConfigSource.java
```diff
@@ -189,6 +189,11 @@ private void reset() {
 		return getPropertyInformations(propertyKey, getConfig());
 	}
 
+	@Override
+	public boolean isSameFile(IPath path) {
```
angelozerr, line 193: It doesn't work on Windows OS since you compare IPathwith java.nio.Path. To fix that you should compare java.io.File. More getOutputConfigFile() can return null.

replace with 

```java
public boolean isSameFile(File file) {
	Path configFile = getOutputConfigFile();
	return configFile != null ? configFile.toFile().equals(file) : false;
}
```

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/IConfigSource.java
```diff
@@ -83,4 +85,12 @@ default int getOrdinal() {
 		return 100;
 	}
 
+	/**
+	 * Returns true if the given path represents the same file as this config source.
+	 *
+	 * @param path the path to check
+	 * @return true if the given path represents the same file as this config source
+	 */
+	boolean isSameFile(IPath path);
```
angelozerr, line 94: `boolean isSameFile(File file);`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
```diff
@@ -187,6 +197,21 @@ private boolean isJavaFile(IFile file) {
...
 		}
 
+		private boolean isConfigSource(IFile file) {
+			IJavaProject project = JavaCore.create(file.getProject());
+			if (project == null) {
+				return false;
+			}
+			for (IConfigSourceProvider provider : ConfigSourceProviderRegistry.getInstance().getProviders()) {
+				for (IConfigSource configSource : provider.getConfigSources(project)) {
+					if (configSource.isSameFile(file.getFullPath())) {
```
angelozerr, line 207: `if (configSource.isSameFile(file.getLocation().toFile())) {`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigDiagnosticsParticipant.java
```diff
@@ -0,0 +1,267 @@
...
+
+				if (nameExpression != null) {
+					name = ((StringLiteral) nameExpression).getLiteralValue();
+				}
+
+				if (name != null && !hasDefaultValue && !doesPropertyHaveValue(name, context)
+						&& !isPropertyIgnored(name)) {
+					String fileUri = context.getUri();
+					IOpenable openable = context.getUtils().resolveCompilationUnit(fileUri);
+					Range range = context.getUtils().toRange(openable, annotation.getStartPosition(),
```
angelozerr, line 189: It should be nice to use the name parameter as range instead of Configproperty annotation range (like validatePropertyDefaultValue does with Expression)

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
```diff
@@ -358,9 +365,8 @@ private void triggerValidationFor(List<String> uris) {
 	}
 
 	public void propertiesChanged(MicroProfilePropertiesChangeEvent event) {
-		if (documents.propertiesChanged(event)) {
-			triggerValidationForAll(event.getProjectURIs());
-		}
+		documents.propertiesChanged(event);
```
angelozerr, line 368: Why do we need this change?

angelozerr: Ok understood. Which fix do you want to do?

datho7561: A part of https://github.com/eclipse/lsp4mp/issues/162; add a new `MicroProfilePropertiesScope` for config properties, and use that to signal that Java files should be revalidated.

datho7561: hold on ; change I made doesn't work

datho7561: I can't seem to figure out what is broken. I'll try on Windows

datho7561: Some tests fail in Eclipse but not when using Maven on Linux

datho7561: On Windows, `microprofile/projectInfo` seems to be failing for me.

datho7561: JavaTextDocument:142, `getNow()` is used, and this appears to always return `null` when getting the project labels in order to filter the documents to only documents in MicroProfile projects. I don't know how I should go about addressing this.

datho7561: Sorry about that. I figured out what was causing the issue.

PR #160, opened 2021-08-03
Remove `QuarkusConfigSourceProvider`

datho7561:
This class provided support for `application.properties` and `application.yaml`. `QuarkusConfigSourceProvider` will be moved to the quarkus-ls project. This follows from #149, where configuration sources can be provided through an extension point. The tests have been updated to use `microprofile-config.properties` instead of `application.properties`, and tests related to Yaml configuration have been removed. As well, the dependency on snakeyaml has been dropped.


angelozerr:
I tested the PR and it looks good. Now the next step is to merge https://github.com/redhat-developer/quarkus-ls/pull/409 (please fix my comments before).

I decided to merge this PR because https://github.com/redhat-developer/quarkus-ls/pull/409 failed with compilation problem.

PR #167, opened 2021-08-23
@ConfigProperty code action when property is not set

### LINKED ISSUE #147: @ConfigProperty code action when property is not set
Once https://github.com/eclipse/lsp4mp/issues/146 will be implemented, we could provide several code actions.

Given this @ConfigProperty:

```java
public class InventoryConfig {

    @ConfigProperty(name="port")
    private int port;

}
```

port will be higlighted as an error, and we could provide the following code actions:

 * `configure in Java `=> it will generate the `, defaultValue="${0}"` attribute
 * `configure in application.properties ` (and other existing properties) => it will generate the `port = ` property
 
For yaml files, I fear that it's not possible to do that because we don't have a Yaml parser with location.

angelozerr:
@ConfigProperty code action when property is not set

Fixes #147

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/CodeActionFactory.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/CodeActionFactory.java
@@ -11,7 +11,7 @@
 * Contributors:
 *     Red Hat Inc. - initial API and implementation
 *******************************************************************************/
-package org.eclipse.lsp4mp.ls.commons;
+package org.eclipse.lsp4mp.jdt.core.java.codeaction;
 
 import java.util.Arrays;
 import java.util.Collections;
@@ -85,8 +85,8 @@ public class CodeActionFactory {
 		TextDocumentEdit textDocumentEdit = new TextDocumentEdit(versionedTextDocumentIdentifier,
 				Collections.singletonList(edit));
 		WorkspaceEdit workspaceEdit = new WorkspaceEdit(Collections.singletonList(Either.forLeft(textDocumentEdit)));
-
 		insertContentAction.setEdit(workspaceEdit);
+		workspaceEdit.setChanges(Collections.emptyMap());
 		return insertContentAction;
 	}
 
@@ -102,12 +102,12 @@ public class CodeActionFactory {
 		TextDocumentEdit textDocumentEdit = new TextDocumentEdit(versionedTextDocumentIdentifier,
 				Collections.singletonList(edit));
 		WorkspaceEdit workspaceEdit = new WorkspaceEdit(Collections.singletonList(Either.forLeft(textDocumentEdit)));
-
 		replaceContentAction.setEdit(workspaceEdit);
+		workspaceEdit.setChanges(Collections.emptyMap());
 		return replaceContentAction;
 	}
 
-	public static boolean isDiagnosticCode(Either<String, Integer> diagnosticCode, String code) {
+	public static boolean isDiagnosticCode(Either<String, Number> diagnosticCode, String code) {
 		if (diagnosticCode == null || diagnosticCode.isRight()) {
 			return false;
 		}
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
@@ -1,0 +1,78 @@
+/*******************************************************************************
+* Copyright (c) 2021 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.jdt.core.java.codeaction;
+
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.Annotation;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.eclipse.lsp4mp.jdt.core.java.corrections.proposal.ChangeCorrectionProposal;
+import org.eclipse.lsp4mp.jdt.core.java.corrections.proposal.InsertAnnotationAttributeProposal;
+
+/**
+ * QuickFix for inserting attribute of a given annotation.
+ *
+ * @author Angelo ZERR
+ *
+ */
+public class InsertAnnotationAttributeQuickFix implements IJavaCodeActionParticipant {
+
+	private static final String CODE_ACTION_LABEL = "Insert ''{0}'' attribute";
+
+	private final String attributeName;
+
+	/**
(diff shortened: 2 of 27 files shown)


angelozerr:
I wrote tests, this PR can be reviewed now.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
```diff
@@ -0,0 +1,78 @@
...
+	 * @param attribute name list of annotation to insert.
+	 */
+	public InsertAnnotationAttributeQuickFix(String attributeName) {
+		this.attributeName = attributeName;
+	}
+
+	@Override
+	public List<? extends CodeAction> getCodeActions(JavaCodeActionContext context, Diagnostic diagnostic,
+			IProgressMonitor monitor) throws CoreException {
+		ASTNode selectedNode = context.getCoveredNode();
```
rgrunber, line 53: If you hover over the undefined annotation member, this works. However, if you quickly click and activate the code action dialog (eg. `ctrl + 1`), `context.getCoveredNode()` will return null. In that case it's probably a 0-length selection which will fail for `getCoveredNode()` because the selection doesn't fully cover anything.

You would generally want to use `getCoveredBy()` since it returns the smallest node that covers the selection.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/CodeActionFactory.java
```diff
@@ -0,0 +1,116 @@
...
+		WorkspaceEdit workspaceEdit = new WorkspaceEdit(Collections.singletonList(Either.forLeft(textDocumentEdit)));
+		insertContentAction.setEdit(workspaceEdit);
+		workspaceEdit.setChanges(Collections.emptyMap());
+		return insertContentAction;
+	}
+
+	public static CodeAction replace(String title, Range range, String replaceText, TextDocumentItem document,
+			Diagnostic diagnostic) {
+		CodeAction insertContentAction = new CodeAction(title);
+		insertContentAction.setKind(CodeActionKind.QuickFix);
```
rgrunber, line 96: `replaceContentAction`

rgrunber:
Overall things look good. I just want to go over `InsertAnnotationAttributeProposal.java` a little more closely. I think we can live with the fact that one must save the properties file or else the same code action will continue to be suggested.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
angelozerr: @rgrunber thanks for the explantation! I fixed all code actions that used `getCoveredNode`. Now I'm using only `getCoveringNode` like you suggested.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/CodeActionFactory.java
angelozerr: fixed.

angelozerr:
>  I just want to go over InsertAnnotationAttributeProposal.java a little more closely. 

Ok I'm waiting for your feedback.

>  think we can live with the fact that one must save the properties file or else the same code action will continue to be suggested.

Yes it can be annoying but I don't know how to save with a standard command the file.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/corrections/proposal/InsertAnnotationAttributeProposal.java
```diff
@@ -0,0 +1,137 @@
...
+		for (int i = 0; i < declaredMethods.length; i++) {
+			IMethodBinding curr = declaredMethods[i];
+			if (!implementedAttribs.contains(curr.getName()) && attributes.contains(curr.getName())) {
+				MemberValuePair pair = ast.newMemberValuePair();
+				pair.setName(ast.newSimpleName(curr.getName()));
+				pair.setValue(newDefaultExpression(ast, curr.getReturnType(), context));
+				listRewriter.insertLast(pair, null);
+
+				addLinkedPosition(rewriter.track(pair.getName()), false, "val_name_" + i); //$NON-NLS-1$
+				addLinkedPosition(rewriter.track(pair.getValue()), false, "val_type_" + i); //$NON-NLS-1$
```
rgrunber, line 107: Just a note about the addLinkedPosition(..) calls. I believe these aren't needed, but I'm fine with keeping them in for potential future enhancements.

In case their use is not clear, linked positions in Eclipse are equivalent to snippet text in VS Code. So when you apply some piece of text and you have some placeholders that you want a user to quickly cycle through (eg. with tab) and modify, this is what is used on the Eclipse side. Notice that with your patch, when you apply the member value insertion, there's no way to cycle between those elements. I think code action doesn't support snippet text the same way completion does, so we can't use this.

However, in the future, if code actions adopt snippet text in the LSP, we could potentially, get the new positions from the underlying proposal and generate a code action that supports snippets (in ChangeUtil).

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/NoValueAssignedToPropertyQuickFix.java
```diff
@@ -0,0 +1,119 @@
...
+	public List<? extends CodeAction> getCodeActions(JavaCodeActionContext context, Diagnostic diagnostic,
+			IProgressMonitor monitor) throws CoreException {
+		List<CodeAction> codeActions = new ArrayList<>();
+		IJavaProject javaProject = context.getJavaProject();
+
+		String lineSeparator = context.getCompilationUnit().findRecommendedLineSeparator();
+		if (lineSeparator == null) {
+			lineSeparator = System.lineSeparator();
+		}
+		String propertyName = getPropertyName(diagnostic, context);
```
rgrunber, line 73: I noticed the code action shows up even when name is empty. (eg. `@ConfigProperty(name= "")`. Is it worth check for empty here to avoid this ? Generating just an `=` in the file seemed a bit odd.

Note that we can't just silence the diagnostic because inserting defaultValue is still valid.

rgrunber (approved):
Overall, this works for me. Only one small issue maybe worth addressing, and I just commented on why linked proposals likely aren't doing anything, but would be good to keep for future compatibility.

Feel free to merge when ready.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/corrections/proposal/InsertAnnotationAttributeProposal.java
angelozerr: Thanks for the good  explanation (I have already linked psoition). You are right we should remove those lines code, it's a copy/paste problem.

I remove those line code.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/NoValueAssignedToPropertyQuickFix.java
angelozerr: @rgrunber after playing with Quarkus application,it seems  that `@ConfigProperty(name="")` is not allowed:

image

It means that we should report an another diagnostic message for this usecase with a code=null instead of NO_VALUE_ASSIGNED_TO_PROPERTY and the problem will be fixed. I created the issue for that https://github.com/eclipse/lsp4mp/issues/176 which will fix your comment below.

rgrunber: Bah, I was hoping it wouldn't be required, and we could get away with just silencing that code action, but fair enough.

PR #168, opened 2021-08-24
added diagnostic for fault-tolerance Asynchronous annotation

### LINKED ISSUE #74: Diagnostics for mp fault-tolerance Asynchronous annotation
https://github.com/eclipse/microprofile-fault-tolerance/blob/master/spec/src/main/asciidoc/asynchronous.asciidoc#asynchronous

Any methods marked with this annotation must return one of:

`java.util.concurrent.Future`
`java.util.concurrent.CompletionStage`
Otherwise, org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException occurs during deployment.

AlexXuChen:
Added diagnostic for microprofile-fault-tolerance `@Asynchronous` annotation.

Closes #74 

Signed-off-by: Alexander Chen alchen@redhat.com


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceDiagnosticsParticipant.java
```diff
@@ -135,6 +138,17 @@ private void validateMethod(MethodDeclaration node, List<Diagnostic> diagnostics
 								diagnostics.add(d);
 							}
 						}
+					} else if (isMatchAnnotation(annotation, ASYNCHRONOUS_ANNOTATION)) {
```
angelozerr, line 141: I will move this code in validateAsynchronousAnnotation (and do the same thing for fallback)

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceDiagnosticsParticipant.java
```diff
@@ -135,6 +138,17 @@ private void validateMethod(MethodDeclaration node, List<Diagnostic> diagnostics
...
+						Type methodReturnType = node.getReturnType2();
+						if (methodReturnType == null || methodReturnType instanceof Future) {
```
rgrunber, line 143: I don't think this is what you want. java.lang.concurrent.Future is what users must ensure used as the method's return type. However, org.eclipse.jdt.core.dom.Type is not going to be an instance of that. It's going to be some object that maybe has a String value that represents "Future". I'd look for some fully qualified name of the type. That's what you want to compare against.

AlexXuChen: Is there another way to cast/declare what the type is for the method?

rgrunber: Yup. You could try :
- Finding all references to `getReturnType2()` in the codebase. This isn't the first time we've used it so we might have other usages where we need to compare the type name against something known. That might give some ideas on how it can be used.
- You could insert a breakpoint right where methodReturnType is assigned and inspect it at runtime. It might give some more ideas about the internal structure and whether it contains what you need.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceDiagnosticsParticipant.java
```diff
@@ -119,26 +125,71 @@ private void validateMethod(MethodDeclaration node, List<Diagnostic> diagnostics
...
+		 * @param annotation  The @Asynchronous annotation 
+		 * @throws JavaModelException
+		 */
+		private void validateAsynchronousAnnotation(MethodDeclaration node, List<Diagnostic> diagnostics,
+				JavaDiagnosticsContext context, NormalAnnotation annotation) throws JavaModelException {
+			Type methodReturnType = node.getReturnType2();
+			String methodReturnTypeString = methodReturnType.resolveBinding().getQualifiedName();
+			System.out.println(methodReturnType);
+			if (methodReturnType == null || 
+					!(methodReturnTypeString == "java.util.concurrent.Future") ||
```
rgrunber, line 182: Using `==` won't work here. For objects, it's best to use their equals() for comparison. Otherwise it's just a reference comparison that's being done.

Try for yourself locally :
`$ echo 'System.out.println(new String("java.util.concurrent.Future") == "java.util.concurrent.Future")' | jshell`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceDiagnosticsParticipant.java
```diff
@@ -119,26 +125,71 @@ private void validateMethod(MethodDeclaration node, List<Diagnostic> diagnostics
...
+		 * @param context     The context, used to create the diagnostics
+		 * @param annotation  The @Asynchronous annotation 
+		 * @throws JavaModelException
+		 */
+		private void validateAsynchronousAnnotation(MethodDeclaration node, List<Diagnostic> diagnostics,
+				JavaDiagnosticsContext context, NormalAnnotation annotation) throws JavaModelException {
+			Type methodReturnType = node.getReturnType2();
+			String methodReturnTypeString = methodReturnType.resolveBinding().getQualifiedName();
+			LOGGER.log(Level.WARNING, "The method return type is " + methodReturnTypeString);
+			if (methodReturnType == null ||
```
rgrunber, line 181: If the method return type is Future or CompletableFuture, then we would still emit diagnostics because of either the 2nd or 3rd conditions being true individually, due to the `||` .

AlexXuChen: Ah sorry thats supposed to be &&

rgrunber: Another quick tip. If you reverse the comparison, so "string".equals(someVar) then you don't need to check for null because the equals will handle it correctly.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceDiagnosticsParticipant.java
```diff
@@ -119,26 +123,70 @@ private void validateMethod(MethodDeclaration node, List<Diagnostic> diagnostics
...
+		 * @param context     The context, used to create the diagnostics
+		 * @param annotation  The @Asynchronous annotation 
+		 * @throws JavaModelException
+		 */
+		private void validateAsynchronousAnnotation(MethodDeclaration node, List<Diagnostic> diagnostics,
+				JavaDiagnosticsContext context, NormalAnnotation annotation) throws JavaModelException {
+			Type methodReturnType = node.getReturnType2();
+			String methodReturnTypeString = methodReturnType.resolveBinding().getQualifiedName();
+			LOGGER.log(Level.INFO, "The method return type is " + methodReturnTypeString);
+			if (methodReturnType.resolveBinding() == null ||
```
rgrunber, line 179: I meant that if `method.resolveBinding()` is null, then you'd get an NPE 2 lines above, where it's also called. But overall this is in better shape now.

I would just confirm that it actually correctly annotates and then we can maybe add some tests.

PR #172, opened 2021-08-27
added 'shouldLanguageServerExitOnShutdown' to ExtendedClientCapabilities

AlexXuChen:
lsp4mp support in ExtendedClientCapabilites for `shouldLanguageServerExitOnShutdown` capability on client side.

References redhat-developer/vscode-microprofile/pull/70

Signed-off-by: Alexander Chen alchen@redhat.com

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/MicroProfileLanguageServer.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/MicroProfileLanguageServer.java
@@ -18,6 +18,9 @@ import static org.eclipse.lsp4j.jsonrpc.CompletableFutures.computeAsync;
 import static org.eclipse.lsp4mp.utils.VersionHelper.getVersion;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/MicroProfileLanguageServer.java
```diff
@@ -68,6 +71,8 @@
 
 	private final MicroProfileExtensionSettings extensionSettings;
 
+	private final ScheduledExecutorService delayer;
```
rgrunber, line 74: I would probably declare this inside shutdown() as I we're only really using it there for now.

rgrunber (commented):
Your commit seems to be missing the Signed-off-by header. Can you make sure to add it ? Should just be `-s` flag of git commit. Also the "lsp4mp support in ExtendedClientCapabilites for shouldLanguageServerExitOnShutdown capability on client side." should probably be in the commit body.

AlexXuChen:
> Your commit seems to be missing the Signed-off-by header. Can you make sure to add it ? Should just be `-s` flag of git commit. Also the "lsp4mp support in ExtendedClientCapabilites for shouldLanguageServerExitOnShutdown capability on client side." should probably be in the commit body.

Does the "Signed-off-by" get added when we merge based on the pr opening comment?

rgrunber (approved):
LGTM. Let me just verify one last thing and I think we can merge.

Issue #163, opened 2021-08-10
Make a SharedSettings class under microprofile-jdt in order to store 


datho7561:
Some settings for lsp4mp need to be accessed in microprofile-jdt. The code would be better structured if these settings were held in a singleton settings object. Make this singleton settings object, and store the existing microprofile-ls settings that are used on the microprofile-jdt side in this settings object.

angelozerr:
The update of settings should trigger validation for Java file too.

PR #174, opened 2021-09-01
added quarkus.http.root-path property to CodeLensURL

AlexXuChen:
Added quarkus.http.root-path property from application.properties to CodeLens URL.

Closes redhat-developer/quarkus-ls/issues/368

*THE CHANGE:
image

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsContext.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsContext.java
@@ -29,6 +29,9 @@ public class JaxRsContext {
 
 	private int serverPort;
 
+	// The quarkus.http.root-path property in application.properties
+	private String rootPath;
+
 	public JaxRsContext() {
 		setServerPort(DEFAULT_PORT);
 	}
@@ -41,6 +44,24 @@ public class JaxRsContext {
 		this.serverPort = serverPort;
 	}
 
+	/**
+	 * Get the quarkus.http.root-path property
+	 *
+	 * @return the rootPath
+	 */
+	public String getRootPath() {
+		return rootPath;
+	}
+
+	/**
+	 * Set the quarkus.http.root-path property
+	 *
+	 * @param rootPath the rootPath to set
+	 */
+	public void setRootPath(String rootPath) {
+		this.rootPath = rootPath;
+	}
+
 	public static JaxRsContext getJaxRsContext(JavaCodeLensContext context) {
 		JaxRsContext jaxRsContext = (JaxRsContext) context.get(CONTEXT_KEY);
 		if (jaxRsContext == null) {
@@ -49,4 +70,19 @@ public class JaxRsContext {
 		}
 		return jaxRsContext;
 	}
+
+	/**
+	 * Create the local base URL
+	 *
+	 * @return the String representation of the project base URL
+	 */
+	public String getLocalBaseURL() {
+		StringBuilder localBaseURL = new StringBuilder("http://localhost:");
+		localBaseURL.append(getServerPort()).toString();
+		if (rootPath != null) {
+			localBaseURL.append(getRootPath());
+		}
+		return localBaseURL.toString();
+	}
+
 }
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
@@ -72,18 +72,18 @@ public class JaxRsCodeLensParticipant implements IJavaCodeLensParticipant {
 	public List<CodeLens> collectCodeLens(JavaCodeLensContext context, IProgressMonitor monitor) throws CoreException {
 		ITypeRoot typeRoot = context.getTypeRoot();
 		IJavaElement[] elements = typeRoot.getChildren();
-		int serverPort = JaxRsContext.getJaxRsContext(context).getServerPort();
+		JaxRsContext jaxrsContext = JaxRsContext.getJaxRsContext(context);
 		IJDTUtils utils = context.getUtils();
 		MicroProfileJavaCodeLensParams params = context.getParams();
-		params.setLocalServerPort(serverPort);
 		List<CodeLens> lenses = new ArrayList<>();
-		collectURLCodeLenses(elements, null, lenses, params, utils, monitor);
+		collectURLCodeLenses(elements, null, lenses, params, jaxrsContext, utils, monitor);
 		return lenses;
 	}
 
 	private static void collectURLCodeLenses(IJavaElement[] elements, String rootPath, Collection<CodeLens> lenses,
-			MicroProfileJavaCodeLensParams params, IJDTUtils utils, IProgressMonitor monitor)
-			throws JavaModelException {
+			MicroProfileJavaCodeLensParams params, JaxRsContext jaxrsContext, IJDTUtils utils, IProgressMonitor monitor)
+			throws JavaModelException, CoreException {
+
 		for (IJavaElement element : elements) {
 			if (monitor.isCanceled()) {
 				return;
@@ -96,10 +96,11 @@ public class JaxRsCodeLensParticipant implements IJavaCodeLensParticipant {
 					// Class is annotated with @Path
 					// Display code lens only if local server is available.
 					if (!params.isCheckServerAvailable()
-							|| isServerAvailable(LOCALHOST, params.getLocalServerPort(), PING_TIMEOUT)) {
+							|| isServerAvailable(LOCALHOST, jaxrsContext.getServerPort(), PING_TIMEOUT)) {
 						// Loop for each method annotated with @Path to generate URL code lens per
 						// method.
-						collectURLCodeLenses(type.getChildren(), pathValue, lenses, params, utils, monitor);
+						collectURLCodeLenses(type.getChildren(), pathValue, lenses, params, jaxrsContext, utils,
+								monitor);
 					}
 				}
 				continue;
@@ -125,7 +126,7 @@ public class JaxRsCodeLensParticipant implements IJavaCodeLensParticipant {
 				// JAX-RS
 				// annotation
 				if (isJaxRsRequestMethod(method) && Flags.isPublic(method.getFlags())) {
-					String baseURL = params.getLocalBaseURL();
+					String baseURL = jaxrsContext.getLocalBaseURL();
 					String openURICommandId = params.getOpenURICommand();
 					CodeLens lens = createURLCodeLens(baseURL, rootPath, openURICommandId, (IMethod) element, utils);
 					if (lens != null) {
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileJavaCodeLensParams.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileJavaCodeLensParams.java
@@ -27,8 +27,6 @@ public class MicroProfileJavaCodeLensParams {
 	private String openURICommand;
 	private boolean checkServerAvailable;
 
-	private transient String localBaseURL;
-
 	public MicroProfileJavaCodeLensParams() {
 
 	}
@@ -90,19 +88,6 @@ public class MicroProfileJavaCodeLensParams {
 	 */
 	public void setLocalServerPort(Integer localServerPort) {
 		this.localServerPort = localServerPort;
-		this.localBaseURL = null;
-	}
-
-	/**
-	 * Returns the local base URL.
-	 *
-	 * @return the local base URL.
-	 */
-	public String getLocalBaseURL() {
-		if (localBaseURL == null) {
-			localBaseURL = new StringBuilder("http://localhost:").append(getLocalServerPort()).toString();
-		}
-		return localBaseURL;
 	}
 
 	/**


angelozerr:
At first LSP4MP should be not linked to quarkus, it means that it should not use `quarkus.http.root-path` in the code. LSP4MP provides a JaxRSContext where you can set some value.

For instance Quarkus provides `quarkus.http.port` property and if you see the LSP4MP code it doesn't use it. The `quarkus.http.port` is retrieved from  quarkus-ls side and update the JaxRscontext#setServerPort with the proper value of `quarkus.http.port`:

https://github.com/redhat-developer/quarkus-ls/blob/d931b173582344dad9357a0ecd1180af0984b718/quarkus.jdt.ext/com.redhat.microprofile.jdt.quarkus/src/main/java/com/redhat/microprofile/jdt/internal/quarkus/jaxrs/java/QuarkusJaxRsCodeLensParticipant.java#L48

After that LSP4MP get the serverPort to build the proper local base URL at https://github.com/eclipse/lsp4mp/blob/cbcc381d2b7240778b7f1422b6f8b67b97759d5a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileJavaCodeLensParams.java#L103

You should try to mimic this strategy. At first I will remove getLocalBaseUrl and getServerPort  from MicroProfileJavaCodeLensParams https://github.com/eclipse/lsp4mp/blob/cbcc381d2b7240778b7f1422b6f8b67b97759d5a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/MicroProfileJavaCodeLensParams.java#L101 because it's parameters (we should not update it) and move those methods in JaxRsContext, the idea is to work with JaxRsContext and not with MicroProfileJavaCodeLensParams to build the URL:

```java
private static void collectURLCodeLenses(IJavaElement[] elements, String rootPath, Collection<CodeLens> lenses,
			JaxRsContext context, IJDTUtils utils, IProgressMonitor monitor)
```

I will add in JaxRsContext a new field with getter/setter `rootPath`. QuarkusJaxRsCodeLensParticipant will update this rootPath with `quarkus.http.root-path`.

The JaxRsContext will use this info to build the URL:

```java
public String getLocalBaseURL() {
   StringBuilder localBaseURL = new StringBuilder("http://localhost:");
  localBaseURL .append(getLocalServerPort()).toString();
   if (rootPath != null) {
      localBaseURL .append(getRootPath());
   }
  return localBaseURL.toString();
}
```

AlexXuChen:
Atm, the codelens url shows the `quarkus.http.root-path`, working on adding JaxRsContext.
Screenshot from 2021-09-02 11-24-28

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsContext.java
```diff
@@ -40,6 +43,20 @@ public int getServerPort() {
 	public void setServerPort(int serverPort) {
 		this.serverPort = serverPort;
 	}
+	
+	/**
+	 * @return the rootPath
```
angelozerr, line 48: Add javadoc

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsContext.java
```diff
@@ -40,6 +43,20 @@ public int getServerPort() {
...
+	/**
+	 * @return the rootPath
+	 */
+	public String getRootPath() {
+		return rootPath;
+	}
+
+	/**
+	 * @param rootPath the rootPath to set
+	 */
```
angelozerr, line 56: Add javadoc

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -127,7 +135,9 @@ private static void collectURLCodeLenses(IJavaElement[] elements, String rootPat
 				if (isJaxRsRequestMethod(method) && Flags.isPublic(method.getFlags())) {
 					String baseURL = params.getLocalBaseURL();
 					String openURICommandId = params.getOpenURICommand();
-					CodeLens lens = createURLCodeLens(baseURL, rootPath, openURICommandId, (IMethod) element, utils);
+					CodeLens lens = rootPathProperty != null ?
```
angelozerr, line 138: Keep existing code:

`CodeLens lens = createURLCodeLens(baseURL, rootPath, openURICommandId, (IMethod) element, utils);`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -72,18 +76,22 @@ public boolean isAdaptedForCodeLens(JavaCodeLensContext context, IProgressMonito
...
 		MicroProfileJavaCodeLensParams params = context.getParams();
 		params.setLocalServerPort(serverPort);
 		List<CodeLens> lenses = new ArrayList<>();
-		collectURLCodeLenses(elements, null, lenses, params, utils, monitor);
+		collectURLCodeLenses(jaxrsContext, elements, null, lenses, params, utils, monitor);
 		return lenses;
 	}
 
-	private static void collectURLCodeLenses(IJavaElement[] elements, String rootPath, Collection<CodeLens> lenses,
+	private static void collectURLCodeLenses(JaxRsContext jaxrsContext, IJavaElement[] elements, String rootPath, Collection<CodeLens> lenses,
```
angelozerr, line 89: Don't use params but only JaxRsContext:

private static void collectURLCodeLenses(IJavaElement[] elements, String rootPath, Collection<CodeLens> lenses,
			JaxRsContext context, IJDTUtils utils, IProgressMonitor monitor)

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -72,18 +76,22 @@ public boolean isAdaptedForCodeLens(JavaCodeLensContext context, IProgressMonito
...
 			MicroProfileJavaCodeLensParams params, IJDTUtils utils, IProgressMonitor monitor)
-			throws JavaModelException {
+			throws JavaModelException, CoreException {
+		
+		String rootPathProperty = jaxrsContext.getRootPath();
```
angelozerr, line 93: Remove this line code.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -99,7 +107,7 @@ private static void collectURLCodeLenses(IJavaElement[] elements, String rootPat
 							|| isServerAvailable(LOCALHOST, params.getLocalServerPort(), PING_TIMEOUT)) {
 						// Loop for each method annotated with @Path to generate URL code lens per
 						// method.
-						collectURLCodeLenses(type.getChildren(), pathValue, lenses, params, utils, monitor);
+						collectURLCodeLenses(jaxrsContext, type.getChildren(), pathValue, lenses, params, utils, monitor);
```
angelozerr, line 110: collectURLCodeLenses(type.getChildren(), pathValue, lenses, context, utils, monitor);

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -127,7 +135,9 @@ private static void collectURLCodeLenses(IJavaElement[] elements, String rootPat
 				if (isJaxRsRequestMethod(method) && Flags.isPublic(method.getFlags())) {
 					String baseURL = params.getLocalBaseURL();
```
angelozerr, line 136: Replace with 

`String baseURL = context.getLocalBaseURL();`

and as I explained in https://github.com/eclipse/lsp4mp/pull/174#issuecomment-911410067 defines in JaxRsContext 

```java
public String getLocalBaseURL() {
   StringBuilder localBaseURL = new StringBuilder("http://localhost:");
  localBaseURL .append(getLocalServerPort()).toString();
   if (rootPath != null) {
      localBaseURL .append(getRootPath());
   }
  return localBaseURL.toString();
}
```

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
AlexXuChen: This handles the null case, so if there is no root-path property, it won't be appended as, localhost:8080/[null]/..

angelozerr: Yes I understand the idea, but see my comments at https://github.com/eclipse/lsp4mp/pull/174#discussion_r701420089

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
AlexXuChen: By removing the params param, there is a lot of refactoring, where there isn't any other method/class that performs similar tasks (e.g. params.isCheckServerAvailable() and params.getOpenURICommand(), both of which can't be moved and shouldn't be copied). Do you still want to remove params here?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsUtils.java
```diff
@@ -66,6 +66,19 @@ public static String getJaxRsPathValue(IAnnotatable annotatable) throws JavaMode
 	public static boolean isJaxRsRequestMethod(IMethod method) throws JavaModelException {
 		return hasAnnotation(method, JAVAX_WS_RS_GET_ANNOTATION);
 	}
+	
+	public static CodeLens createURLCodeLens(String baseURL, String route, String rootPath, String openURICommandId, IMethod method,
```
angelozerr, line 70: Don't change this method by adding route as parameter. The route should be included inside baseURL.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
angelozerr: Ok let's keep params only for  params.isCheckServerAvailable() and params.getOpenURICommand(). But we should have JaxRSContext#getLocalBaseURL().

Please move JaxRSContext (from the first parameter) after the params.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -72,18 +72,20 @@ public boolean isAdaptedForCodeLens(JavaCodeLensContext context, IProgressMonito
 	public List<CodeLens> collectCodeLens(JavaCodeLensContext context, IProgressMonitor monitor) throws CoreException {
 		ITypeRoot typeRoot = context.getTypeRoot();
 		IJavaElement[] elements = typeRoot.getChildren();
-		int serverPort = JaxRsContext.getJaxRsContext(context).getServerPort();
+		JaxRsContext jaxrsContext = JaxRsContext.getJaxRsContext(context);
+		int serverPort = jaxrsContext.getServerPort();
 		IJDTUtils utils = context.getUtils();
 		MicroProfileJavaCodeLensParams params = context.getParams();
 		params.setLocalServerPort(serverPort);
```
angelozerr, line 79: Remove setLocalServerPort from MicroProfileJavaCodeLensParams

AlexXuChen: I think we should keep the setter. Also MicroProfileDelegateCommandHandlerForJava.java line 228 uses it.

angelozerr: Ok keep like this, we will remove it when we will implement https://github.com/eclipse/lsp4mp/issues/163, but please remove code:

```java
int serverPort = jaxrsContext.getServerPort();
MicroProfileJavaCodeLensParams params = context.getParams();
params.setLocalServerPort(serverPort);
```

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsContext.java
```diff
@@ -49,4 +68,18 @@ public static JaxRsContext getJaxRsContext(JavaCodeLensContext context) {
...
+	
+	/** Create the local base URL 
+	 * 
+	 * @return the String representation of the project base URL
+	 */
+	public String getLocalBaseURL() {
+	  StringBuilder localBaseURL = new StringBuilder("http://localhost:");
+	  localBaseURL.append(getServerPort()).toString();
+	   if (rootPath != null) {
+	      localBaseURL.append(getRootPath());
```
angelozerr, line 80: What is the behavior of quarkus if root path doesn't starts with '/' ?

AlexXuChen: It will throw an error on build:
Screenshot from 2021-09-07 13-23-30

angelozerr: Ok thanks for the feedback. Could you create an issue to validate this property on application properties file please

AlexXuChen: Are we good to close this one then?

angelozerr: Yes let's keep like this.

angelozerr:
@AlexXuChen this PR doesn't fix all issues from redhat-developer/quarkus-ls/issues/368.

I re-read it and it seems `@ApplicationPath` Java annotation play the same role than the quarkus property `quarkus.http.root-path`. As `@ApplicationPath` is a part of MicroProfile spec; it should be managed on LSP4MP side.

According to  @clochardpagan comments:

 * there is no option to work with host differing from localhost. That's a problem.
 * there is no option to work with https:// at all. Bad, but can be fixed via SSL termination or/and proxy.

To fix this problem an idea is to provide a setting to declare a base URL for that. Or perhaps we could declare an array of baseURL to generate several Codelens URL for serveral server. @fbricon @rgrunber what do you think about that?

@AlexXuChen those issues could be done in an another PRs, but please change your commit which closes the issue 368, because this PR doesn't covers all issues.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/jaxrs/JaxRsContext.java
AlexXuChen: If we're good to close, can we merge this and I'll open another pr to handle the 2 cases:
- there is no option to work with host differing from localhost. That's a problem.
- there is no option to work with https:// at all. Bad, but can be fixed via SSL termination or/and proxy.

rgrunber (commented):
Could `localBaseURL` that's used in `MicroProfileJavaCodeLensParams` be removed ? I don't think it's used anywhere now that `JaxRsContext` took over the same functionality.

There's also extraneous whitespace throughout the files. Just need to remove them.

AlexXuChen:
> Could `localBaseURL` that's used in `MicroProfileJavaCodeLensParams` be removed ? I don't think it's used anywhere now that `JaxRsContext` took over the same functionality.

Originally `MicroProfileJavaCodeLensParams` was used before the `JaxRsContext` was required, and that was the only use of `localBaseURL` from the params. So yes, I believe it's okay to remove since it is unused.

rgrunber:
Yeah, there seems to be a bit of overlap between the MicroProfile params and the JaxRsContext. Maybe there's some parameters that need to reside in params for a little while before they get transfered over to the JaxRsContent but the localBaseURL stood out as the single getter in MicroProfile params wans't being called anywhere after you switched over and looked pretty much the same as the new one in JaxRsContext.

rgrunber (approved):
I've tried this out (with some modifications on quarkus-ls) and it seems to be working as expected.

PR #178, opened 2021-09-13
Run LSP4J callback methods in separate thread

### LINKED ISSUE #177: Deadlock when client is sending burst of request
In IJ Quarkus, when the IDE starts, if several Java files are opened, we will send a didOpen request for each file. The server will send back a getProjectLabels request for each file and this causes a deadlock. The reason is that each LSP request method should not run in the calling thread because it prevents the server to read the input stream of LSP4J is single thread per connection.

Copied from https://github.com/redhat-developer/quarkus-ls/issues/338

jeffmaury:
Fixes #177

Signed-off-by: Jeff MAURY <jmaury@redhat.com>


angelozerr:
Many thanks @jeffmaury! I will review it ASAP.

Since JDT-LS uses the same idea with ParentProcessWatcher like MP LS, @rgrunber what do think about that?

rgrunber:
> Many thanks @jeffmaury! I will review it ASAP.
> 
> Since JDT-LS uses the same idea with ParentProcessWatcher like MP LS, @rgrunber what do think about that?

This isn't reproducible with any other language server. As you mentioned, we even do this for JDT-LS. Could we add some kind of setting for this so clients can enable/disable as needed, given that we don't immediately know what the implications may be on every client ?

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/ls/MicroProfileLSTest.java
```diff
@@ -0,0 +1,129 @@
...
+ * Copyright (c) 2021 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.eclipse.lsp4mp.ls;
```
angelozerr, line 11: Please move all your classesof your test in `org.eclipse.lsp4mp.ls.deadlock` package for instance.

angelozerr:
@jeffmaury could you squash your 3 commits in one please.

jeffmaury:
> 
> 
> @jeffmaury could you squash your 3 commits in one please.

GitHub UI allows you to do so when merging

angelozerr:
I testes your PR with vscode-microprofile and there is no problem (because I have not activated runAsync system property). I try your test by removing runAsync  and indeed we see the problem with deadlock.

For the moment we disable runAsync for vscode-microprofile. If we see some deadlock troubles in vscode-microprofile we will activate it. Today we have just seen the deadlock problem with IJ.

Thanks so much for your work @jeffmaury!

PR #180, opened 2021-09-16
[MP Config 2.0] Support for config profiles

### LINKED ISSUE #78: [MP Config 2.0] Support for config profiles
https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configprofile.asciidoc

The property `mp.config.profile` is used to specify the profile being used.

Config Profiles can work on the property level and source file level. 

Ie.

Property level: 

A profile property matching the active profile name overrides a regular property in the same config source.

```
%dev.vehicle.name=car
%live.vehicle.name=train
%testing.vehicle.name=bike
vehicle.name=lorry
```

Config source level:

A config source matching the active profile name will be loaded on top of `microprofile-config.properties`

```
META-INF\microprofile-config.properties
META-INF\microprofile-config-dev.properties
META-INF\microprofile-config-prod.properties
META-INF\microprofile-config-testing.properties
```

angelozerr:
[MP Config 2.0] Support for config profiles

Fixes #78


angelozerr:
As I'm breaking API for config source, I create a PR for Quarkus side at https://github.com/redhat-developer/quarkus-ls/pull/418

rgrunber:
Ok, so I'm able to confirm that if I define multiple config property files of form "microprofile-config-${profile}.properties", then all the different properties are displayed on a hover in a Java source file. Is there any other functionality this adds ? Setting mp.config.profile in any property file doesn't do anything for me.

rgrunber (approved):
Was able to get the main user-visible feature to work so i'm fine with merging. If there's other issues, we can always file separate bugs.

angelozerr:
The test which is important is to delete add properties files and check hover or diagnostic for configproperty name attributs is working

angelozerr:
Thanksa lot @rgrunber for your review!

Issue #77, opened 2020-09-15
Diagnostics for invalid annotation parameter values


rzgry:
Several mp fault tolerance annotations will throw a `FaultToleranceDefinitionException` if configured incorrectly.  While I noticed this for fault tolerance, probably would be good to design this in a way that could be re-used for other mp-specs annotations.

A common scenario is parameters that have a min/max value

Ex. circuit breaker failure ratio must be between 0-1 https://github.com/eclipse/microprofile-fault-tolerance/blob/master/api/src/main/java/org/eclipse/microprofile/faulttolerance/CircuitBreaker.java#L149

or requestVolumeThreshold  must be >= 1 https://github.com/eclipse/microprofile-fault-tolerance/blob/master/api/src/main/java/org/eclipse/microprofile/faulttolerance/CircuitBreaker.java#L136

Can look at the invalid parameters tests for the different cases annotations are invalid https://github.com/eclipse/microprofile-fault-tolerance/tree/5adfe495f0ad1fe8f52a443781b7509b01a92f15/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/invalidParameters

datho7561:
Here is some info to help with implementing this issue:

Here is the current class we use to validate MicroProfile Fault Tolerance annotations: https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceDiagnosticsParticipant.java

Here is a utils library that we have for helping work with annotations in JDT: https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java

AlexXuChen:
I can work on this issue

rgrunber:
I'm generally ok, with adding the annotation support, but where I'd like to have some automation or generalization is in the validation of the parameters themselves. It seems overkill to have to hard-code specific parameter restrictions. Granted, we've done pretty much the same for #90 & https://github.com/redhat-developer/quarkus-ls/issues/378

I wonder if we could have some general definition file that specifies a format like :
```
@CircuitBreaker,failureRatio,double,0-1
@CircuitBreaker,successThreshold,int,>=1
...
...
```

[...]

PR #184, opened 2021-09-24
added diagnostic support for Bulkhead annotation value and waitingTaskQueue

AlexXuChen:
Added diagnostic support for `@Bulkhead` annotation of `value` and `waitingTaskQueue` member(s)

References #77


AlexXuChen:
Fixed in #189

PR #185, opened 2021-09-24
added diagnostic support for Timeout annotation value

AlexXuChen:
Added diagnostic support for `@Timeout` annotation for `value` member

References #77


AlexXuChen:
Fixed in #189

PR #191, opened 2021-09-28
added diagnostics support for `@Retry` annotation member values

AlexXuChen:
Added diagnostics support for `@Retry` annotation member values, for non-negative values and `delay` exceeding `maxDuration`

References #77 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
@@ -19,13 +19,23 @@ import static org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants.UNI_TYPE_U
 import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.getAnnotationMemberValueExpression;
 import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.isMatchAnnotation;
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.ASYNCHRONOUS_ANNOTATION;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.DELAY_RETRY_ANNOTATION_MEMBER;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.DELAY_UNIT_RETRY_ANNOTATION_MEMBER;
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.DIAGNOSTIC_SOURCE;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.DURATION_UNIT_RETRY_ANNOTATION_MEMBER;
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.FALLBACK_ANNOTATION;
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.FALLBACK_METHOD_FALLBACK_ANNOTATION_MEMBER;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.JITTER_DELAY_UNIT_RETRY_ANNOTATION_MEMBER;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.JITTER_RETRY_ANNOTATION_MEMBER;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.MAX_DURATION_RETRY_ANNOTATION_MEMBER;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.MicroProfileFaultToleranceConstants.RETRY_ANNOTATION;
+import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceErrorCode.DELAY_EXCEEDS_MAX_DURATION;
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceErrorCode.FALLBACK_METHOD_DOES_NOT_EXIST;
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceErrorCode.FAULT_TOLERANCE_DEFINITION_EXCEPTION;
 
 import java.text.MessageFormat;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -56,14 +66,16 @@ import org.eclipse.lsp4mp.jdt.core.java.validators.JavaASTValidator;
 import org.eclipse.lsp4mp.jdt.core.utils.JDTTypeUtils;
 
 /**
- * Collects diagnostics related to the <code>@Fallback</code> and
- * <code>@Asynchronous</code> annotations.
+ * Collects diagnostics related to the <code>@Fallback</code>,
+ * <code>@Asynchronous</code> and <code>@Retry</code> annotations.
  */
 public class MicroProfileFaultToleranceASTValidator extends JavaASTValidator {
 
 	private static final String FALLBACK_ERROR_MESSAGE = "The referenced fallback method ''{0}'' does not exist.";
 
 	private static final String ASYNCHRONOUS_ERROR_MESSAGE = "The annotated method ''{0}'' with @Asynchronous should return an object of type {1}.";
+
+	private static final String RETRY_ERROR_MESSAGE = "The `delay` member value must be less than the `maxDuration` member value.";
 
 	private final Map<TypeDeclaration, Set<String>> methodsCache;
 
@@ -83,7 +95,8 @@ public class MicroProfileFaultToleranceASTValidator extends JavaASTValidator {
 			throws CoreException {
 		IJavaProject javaProject = context.getJavaProject();
 		boolean adapted = JDTTypeUtils.findType(javaProject, FALLBACK_ANNOTATION) != null
-				|| JDTTypeUtils.findType(javaProject, ASYNCHRONOUS_ANNOTATION) != null;
+				|| JDTTypeUtils.findType(javaProject, ASYNCHRONOUS_ANNOTATION) != null
+				|| JDTTypeUtils.findType(javaProject, RETRY_ANNOTATION) != null;
 		if (adapted) {
 			addAllowedReturnTypeForAsynchronousAnnotation(javaProject, UNI_TYPE_UTILITY);
 		}
@@ -123,7 +136,12 @@ public class MicroProfileFaultToleranceASTValidator extends JavaASTValidator {
 					} catch (JavaModelException e) {
 						LOGGER.log(Level.WARNING, "An exception occurred when attempting to validate the annotation");
 					}
-					break;
+				} else if (isMatchAnnotation(annotation, RETRY_ANNOTATION)) {
+					try {
+						validateRetryAnnotation((NormalAnnotation) modifier);
+					} catch (JavaModelException e) {
+						LOGGER.log(Level.WARNING, "An exception occurred when attempting to validate the annotation");
+					}
 				}
 			}
 		}
@@ -145,11 +163,11 @@ public class MicroProfileFaultToleranceASTValidator extends JavaASTValidator {
 			if (modifier instanceof Annotation) {
 				Annotation annotation = (Annotation) modifier;
 				if (isMatchAnnotation(annotation, FALLBACK_ANNOTATION)) {
-					NormalAnnotation normAnnotation = (NormalAnnotation) modifier;
-					validateFallbackAnnotation(node, normAnnotation);
+					validateFallbackAnnotation(node, (NormalAnnotation) modifier);
 				} else if (isMatchAnnotation(annotation, ASYNCHRONOUS_ANNOTATION)) {
-					MarkerAnnotation markAnnotation = (MarkerAnnotation) modifier;
-					validateAsynchronousAnnotation(node, markAnnotation);
+					validateAsynchronousAnnotation(node, (MarkerAnnotation) modifier);
+				} else if (isMatchAnnotation(annotation, RETRY_ANNOTATION)) {
+					validateRetryAnnotation((NormalAnnotation) modifier);
 				}
 			}
 		}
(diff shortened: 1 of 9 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -204,6 +230,29 @@ private void validateAsynchronousAnnotation(MethodDeclaration node, MarkerAnnota
...
+	 * @param annotation The @Retry annotation
+	 * @throws JavaModelException
+	 */
+	private void validateRetryAnnotation(MethodDeclaration node, NormalAnnotation annotation)
+			throws JavaModelException {
+		Expression delayExpr = getAnnotationMemberValueExpression(annotation, DELAY_RETRY_ANNOTATION_MEMBER);
+		Expression maxDurationExpr = getAnnotationMemberValueExpression(annotation,
+				MAX_DURATION_RETRY_ANNOTATION_MEMBER);
+		if (delayExpr != null && maxDurationExpr != null) {
+			int delayValue = Integer.parseInt(delayExpr.toString());
```
rgrunber, line 247: I don't think the way to do it is by using toString() directly. An Expression can be a StringLiteral, or a NumberLiteral, etc. I think you need to check that it's the one you expect, and then get the value from that.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -115,14 +123,24 @@ public boolean visit(TypeDeclaration type) {
 			if (modifier instanceof Annotation) {
 				Annotation annotation = (Annotation) modifier;
 				if (isMatchAnnotation(annotation, ASYNCHRONOUS_ANNOTATION)) {
-					try {
-						MethodDeclaration[] methods = type.getMethods();
-						for (MethodDeclaration node : methods) {
-							validateAsynchronousAnnotation(node, (MarkerAnnotation) modifier);
+					BiConsumer<MethodDeclaration, Annotation> validateAsynchronous = (scanType, scanAnnotation) -> {
```
rgrunber, line 126: Is there any reason to do it this way ? Why not just call both methods here.

rgrunber:
It works, but there's a small problem I noticed. It only makes sense to compare the delay and maxDuration directly when their time units (delayUnit & durationUnit) match. If they don't, you may need to do some additional math to be sure.

For example If delay = 5, maxDuration = 3, delayUnit = ChronoUnit.SECONDS, durationUnit = ChronoUnit.MILLENNIA, validation would currently fail, but we should be fine.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
AlexXuChen: The original intention was to reduce the repetitive code for iterating through the methods when the annotation was declared in the class. However, given `@Retry` doesn't interact with methods like `@Asynchronous` does, I don't believe I need it.

rgrunber: I see your point. It saves having to write out the for-loop in each condition, but you still end up writing out the method declaration that gets passed it. I just found the code a bit more confusing because you have to follow the calls to ultimately figure out that accept(T, U) of the BiConsumer finally does the validation.

I thought just calling directly would probably be easier to read.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
AlexXuChen: Without toString(), parseInt won't work since Java doesn't seem to recognize an Expression as a StringLiteral. Should I use a cast instead?

rgrunber: Actually, I think the value would be a NumberLiteral, so you would have to check for that. Either way, if you insert a breakpoint here and check the type of Expression, it should tell you. Then you just have to find what method to use in order to get at the value.

AlexXuChen: It seems when it's a negative number, it's classified as a PrefixExpression, and when it's a positive number, it's considered a NumberLiteral. However in the Java code, it is always assigned as an Expression, which will throw a Syntax error when using any operator that accepts a number.

rgrunber: But we don't need to support negative numbers for this particular validation. In fact if it were a negative number, you could safely fail the validation. Did you check whether Expression has any methods that might be of use in getting to something you can use ?

AlexXuChen: Expression has `resolveConstantExpressionValue()`, which should work out for what we want to implement.

For supporting negative values, if we have 
```java
@Retry(delay = -1000, maxDuration = -5000)
```
we should throw the negative diagnostic, but should we also throw the `DELAY_EXCEEDS_MAX_DURATION` error (which is supported atm)?

rgrunber: We shouldn't throw the delay exceeding max duration is only really defined and meaningful when we're deal with positive numbers. If any of those values are negative, then the first thing to address is the negative value.

AlexXuChen:
According to the docstring for jitter(https://github.com/eclipse/microprofile-fault-tolerance/blob/800363fe002701bb2e31ba413f21e5a36af56b41/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java#L120), I believe that the jitter can randomly add to/subtract from the delay. That means if the delay + jitter exceeds maxDuration, then it should be invalid. However, there doesn't seem to be a test case for this.

Should this case be considered in validation @rgrunber?

rgrunber:
@angelozerr , looks like RangeExpression.parse(String) can't deal with the `-` from `-1` . We need such values as sometimes it can indicate "infinite wait" for things like a delay.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -121,9 +132,14 @@ public boolean visit(TypeDeclaration type) {
 							validateAsynchronousAnnotation(node, (MarkerAnnotation) modifier);
 						}
 					} catch (JavaModelException e) {
-						LOGGER.log(Level.WARNING, "An exception occurred when attempting to validate the annotation");
+						e.printStackTrace();
```
rgrunber, line 135: Avoid just leaving these as `e.printStrackTrace()` . When you catch an exception, just ask yourself whether you expect it to ever fail, and if so, whether you can recover, or simply reporting the error is the best that can be done.

In this case, it's like you just want to report the error using `LOGGER.log(..) as it was before on the left.

rgrunber (commented):
I think we just need to wait on a fix for lsp4mp to handle negative values in the range.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -204,6 +220,37 @@ private void validateAsynchronousAnnotation(MethodDeclaration node, MarkerAnnota
...
+				MAX_DURATION_RETRY_ANNOTATION_MEMBER);
+		if (delayExpr != null && maxDurationExpr != null) {
+			Expression delayUnitExpr = getAnnotationMemberValueExpression(annotation,
+					DELAY_UNIT_RETRY_ANNOTATION_MEMBER);
+			Expression durationUnitExpr = getAnnotationMemberValueExpression(annotation,
+					DURATION_UNIT_RETRY_ANNOTATION_MEMBER);
+			int delayNum = (int) delayExpr.resolveConstantExpressionValue();
+			int maxDurationNum = (int) maxDurationExpr.resolveConstantExpressionValue();
+			Duration delayValue = delayUnitExpr != null
+					? Duration.of(delayNum, ChronoUnit.valueOf(delayUnitExpr.toString().split("\\.")[1]))
```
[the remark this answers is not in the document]
AlexXuChen: This would make the `xExpr != null` repetitive, is that alright?

rgrunber:
Can you try adding :

```diff
diff --git a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/RangeExpression.java b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/RangeExpression.java
index 9b70b92..82cd29a 100644
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/RangeExpression.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/RangeExpression.java
@@ -152,7 +152,7 @@ public class RangeExpression {
                                valueAsString.append(c);
                                break;
                        default:
-                               if (Character.isDigit(c)) {
+                               if (Character.isDigit(c) || '-' == c) {^M
                                        valueAsString.append(c);
                                } else {
                                        unexpectedToken(c, i, expression);
```
and see if that solves the issue ?

AlexXuChen:
> and see if that solves the issue ?

This is a solution, but I also added the `case '-'`: https://github.com/eclipse/lsp4mp/pull/191/files#diff-3e4d6d0d761c9ffc6f9f9d20379ec87c91773ab5778f245110280ad427691a5fR154 which seems to handle the `-`
Screenshot from 2021-09-30 16-19-57
Screenshot from 2021-09-30 16-20-06

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationValidatorTest.java
```diff
@@ -47,6 +47,13 @@ public void testGreaterThanOrEqual2() throws RangeExpressionException {
 		assertValidation("[0", "-1", "The value `-1` must be greater than or equal to `0`.");
 	}
 
+	@Test
+	public void testGreaterThanOrEqual3() throws RangeExpressionException {
+		assertValidation("-1", "0", null);
```
angelozerr, line 52: It misses the case with assertValidation("-1", "-1", null);

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationValidatorTest.java
```diff
@@ -47,6 +47,13 @@ public void testGreaterThanOrEqual2() throws RangeExpressionException {
 		assertValidation("[0", "-1", "The value `-1` must be greater than or equal to `0`.");
 	}
 
+	@Test
+	public void testGreaterThanOrEqual3() throws RangeExpressionException {
```
angelozerr, line 51: I will rename with testGreaterThanOrEqualWithMinusValue

angelozerr:
LGTM

angelozerr:
Great job @AlexXuChen !

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -204,6 +222,53 @@ private void validateAsynchronousAnnotation(MethodDeclaration node, MarkerAnnota
...
+
+			Duration delayValue = delayUnitExpr != null ? Duration.of(delayNum, ChronoUnit.valueOf(delayUnit))
+					: Duration.of(delayNum, ChronoUnit.MILLIS);
+			Duration maxDurationValue = durationUnitExpr != null
+					? Duration.of(maxDurationNum, ChronoUnit.valueOf(maxDurationUnit))
+					: Duration.of(maxDurationNum, ChronoUnit.MILLIS);
+			Duration jitterValue = jitterDelayUnitExpr != null
+					? Duration.of(jitterNum, ChronoUnit.valueOf(jitterDelayUnit))
+					: Duration.of(jitterNum, ChronoUnit.MILLIS);
+			Duration maxDelayValue = delayValue.plus(jitterValue);
```
rgrunber, line 263: Ok, jitter it is. It isn't just added to the delay, but also subtracted. It adds some random delay to the minium value. Have a look at https://github.com/eclipse/microprofile-fault-tolerance/blob/800363fe002701bb2e31ba413f21e5a36af56b41/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java#L113 .

The reason I hesitated is because if you take a delay of 3, with a jitter of 2, and maxDuration of 4, you'll have delays between [1,5] . Some of the delays will fall below 4, but others above. Should we all this an error ? Maybe a warning ?

rgrunber (changes requested):
Try the following :

```
@Retry(
        delay = 8,
        maxDuration = 2,
        delayUnit = java.time.temporal.ChronoUnit.HOURS,
        durationUnit = ChronoUnit.HOURS
    )
```
Basically, you need to deal with fully qualified names.

A fully qualified name is made of 2 components : The qualifier (eg. java.time.temporal.ChronoUnit, or just ChronoUnit) and the name (eg. HOURS).

rgrunber:
You can also declare : `import static java.time.temporal.ChronoUnit.HOURS;` and then use HOURS directly, in which case it won't be a QualifiedName, but just a SimpleName. This and the above will break your validation.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
AlexXuChen: Currently the error message reports "The `delay` member value must be less than the `maxDuration` member value.", but to illustrate this, I can mention "the sum of `delay` and `jitter`...". 

According to https://github.com/eclipse/microprofile-fault-tolerance/blob/800363fe002701bb2e31ba413f21e5a36af56b41/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java#L113, the negative effective delays have a lower bound of 0, which means the maximum effective delay should be the only one we need to validate.

PR #192, opened 2021-10-01
added diagnostic for `@ConfigProperty` name member as empty string

### LINKED ISSUE #176: Report diagnostic for @ConfigProperty(name="")
Given a class with @ConfigProperty name attribute is empty:

```java
public class Foo {

  @ConfigProperty(name="")
  private String bar;

}
```

The Quarkus application throws this error :

image

We should report a diagnostic when name="" and it willfix too https://github.com/eclipse/lsp4mp/pull/167#discussion_r705496691

AlexXuChen:
Added diagnostic for `@ConfigProperty` name member as empty string to handle `name=""` case.

Fixes #176 


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
```diff
@@ -187,7 +189,11 @@ private void validatePropertyHasValue(Annotation annotation, Expression defaultV
 				name = MicroProfileConfigPropertyProvider.getPropertyName(name, currentPrefix);
 			}
 
-			if (name != null && !hasDefaultValue && !doesPropertyHaveValue(name, getContext())
+			if (name != null && name.isEmpty()) {
```
angelozerr, line 192: I will write just one if (name != null) and after that write if (name.isEmpty()))

rgrunber (approved):
Looks good. I think we can merge this.

PR #193, opened 2021-10-04
fixed `ChronoUnit` static import case for `@Retry` annotation unit

AlexXuChen:
Fixed `ChronoUnit` static import case for `@Retry` annotation unit(s)

Fixes https://github.com/eclipse/lsp4mp/pull/191#pullrequestreview-769314742


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -240,26 +242,41 @@ private void validateRetryAnnotation(NormalAnnotation annotation) throws JavaMod
...
+
+			String jitterUnit = null;
+			if (durationUnitExpr != null) {
+				SimpleName jitterUnitName = jitterUnitExpr instanceof SimpleName ? (SimpleName) jitterUnitExpr
+						: ((QualifiedName) jitterUnitExpr).getName();
+				jitterUnit = jitterUnitName.getIdentifier();
+			}
+			Duration jitterValue = jitterUnit != null ? Duration.of(jitterNum, ChronoUnit.valueOf(jitterUnit))
 					: Duration.of(jitterNum, ChronoUnit.MILLIS);
+
```
rgrunber, line 279: https://github.com/eclipse/lsp4mp/pull/191#discussion_r720494385 .

The effective delay is a range, so you'll end up also having a minDelayValue when subtracting the jitterValue from the delayValue.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -240,26 +242,41 @@ private void validateRetryAnnotation(NormalAnnotation annotation) throws JavaMod
...
-			Duration maxDurationValue = durationUnitExpr != null
-					? Duration.of(maxDurationNum, ChronoUnit.valueOf(maxDurationUnit))
+
+			String durationUnit = null;
+			if (durationUnitExpr != null) {
+				SimpleName durationUnitName = durationUnitExpr instanceof SimpleName ? (SimpleName) durationUnitExpr
+						: ((QualifiedName) durationUnitExpr).getName();
+				durationUnit = durationUnitName.getIdentifier();
+			}
+			Duration maxDurationValue = durationUnit != null
```
rgrunber, line 267: I would format your trinary expressions the way you did this one.

rgrunber (changes requested):
```
@Retry(
        maxDuration = 5,
        delay = 5,
        delayUnit = ChronoUnit.HOURS,
        durationUnit = ChronoUnit.HOURS
    )
```
doesn't work for me. No diagnostic displayed.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
AlexXuChen: I believe the minDelayValue would have any effect on the error, correct? According to mp-fault-tolerance, the minimum effective delay is bound by 0.

rgrunber: Bound by 0 when the jitter would force it to be less than 0 (since such a value would not make sense). For example, if the delay were 2 and the jitter were 5, the range would be `[2-5, 2+5]=[-3,7]`. However, because of that condition, the range becomes [0,7].

AlexXuChen: Are there any diagnostics that should be thrown for this?

rgrunber: Would you be fine with just converting the diagnostic to a warning and simply being something like :

"The effective delay may exceed the maximum duration."

rgrunber: Whoops, I'm being silly. You're right, this is not an issue. If delay + jitter < maxDuration we don't have to worry about the ranges overlapping.

rgrunber:
One last thing I noticed. Try setting one of the unit values to ChronoUnit.YEARS. You should notice the following error :

`java.time.temporal.UnsupportedTemporalTypeException: Unit must not have an estimated duration`

It looks like `Duration.of(long, TemporalUnit) does not support any value larger than DAYS. You can see this in the source code for Duration.of(...) -> Duration.plus(..)

```
if (unit.isDurationEstimated()) {
    throw new UnsupportedTemporalTypeException("Unit must not have an estimated duration");
}
```

Can we just skip the calculation for now if units larger than DAYS are used or is there a way around this ?

AlexXuChen:
> Can we just skip the calculation for now if units larger than DAYS are used or is there a way around this ?

After some digging, I found that most of the Duration methods that have Temporal params do the same thing
e.g. for plus, it only supports up to days: https://stackoverflow.com/questions/39907925/why-instant-does-not-support-operations-with-chronounit-years

From the looks of it, this is because the implementation converts the representation to seconds and compares, and this conversion stops after DAYS.

There are two possible fixes here:
1. We make a wrapper for Duration.of and handle YEARS and onward
2. We throw an error for YEARS and onward and skip the calculation (as you mentioned)
Likely the latter solution is more practical since I can't imagine why you would want a method retry to be delayed by a x number of years.

rgrunber:
> > Can we just skip the calculation for now if units larger than DAYS are used or is there a way around this ?
> 
> After some digging, I found that most of the Duration methods that have Temporal params do the same thing e.g. for plus, it only supports up to days: https://stackoverflow.com/questions/39907925/why-instant-does-not-support-operations-with-chronounit-years
> 
> From the looks of it, this is because the implementation converts the representation to seconds and compares, and this conversion stops after DAYS.
> 
> There are two possible fixes here:
> 
>     1. We make a wrapper for Duration.of and handle YEARS and onward
> 
>     2. We throw an error for YEARS and onward and skip the calculation (as you mentioned)
> 
> 
> Likely the latter solution is more practical since I can't imagine why you would want a method retry to be delayed by a x number of years. Thoughts @rgrunber?

I don't think we need to throw an error, beacuse technically we're in the wrong for not supporting a valid use case. The spec says ChronoUnit and that means all values, but we can certainly opt to not support something for now, if it's too much of a pain. You can probably add a simple check for now that simply skips validation if any of the ChronoUnit values are above DAYS.

AlexXuChen:
> I don't think we need to throw an error, beacuse technically we're in the wrong for not supporting a valid use case. The spec says ChronoUnit and that means all values, but we can certainly opt to not support something for now, if it's too much of a pain. You can probably add a simple check for now that simply skips validation if any of the ChronoUnit values are above DAYS.

I don't see any indication on the client side when the `java.time.temporal.UnsupportedTemporalTypeException` is thrown. For our "fix", should we just simply ignore the calculation when we have an UnsupportedTemporalTypeException with no indicator on the client side?

rgrunber:
> I don't see any indication on the client side when the `java.time.temporal.UnsupportedTemporalTypeException` is thrown. For our "fix", should we just simply ignore the calculation when we have an UnsupportedTemporalTypeException with no indicator on the client side? Otherwise a wrapper is our best bet.

That's because the error is caught by https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/MultiASTVisitor.java#L120 when run at https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java#L255-L262 .

It'll end up in the logs, so I would just check if any of the ChronoUnit values use a value greater than DAYS and skip the validation.

rgrunber:
`ChronoUnit.valueOf(String).isDateBased()` should also work.

AlexXuChen:
> `ChronoUnit.valueOf(String).isDateBased()` should also work.

This would make DAYS return true as well, since DAYS is date based, not time based.

AlexXuChen:
> so it likely fails for DAYS also. I just happened to try YEARS. Please check whether DAYS causes the issue as well and if so we can simply use one of the helper methods for the enumeration.

If you take
```
ChronoUnit.DAYS.isDurationEstimated();
```
and
```
ChronoUnit.YEARS.isDurationEstimated();
```
both result in true, however, if you use DAYS in the Retry annotation, we still get a warning (as desired):
Screenshot from 2021-10-05 16-24-37

I think the main point right now is to decide which strategy we want to use to ignore non-supported Temporal arguments for Duration.

AlexXuChen:
For Duration.plus,
```java
public Duration plus(long amountToAdd, TemporalUnit unit) {
        Objects.requireNonNull(unit, "unit");
        if (unit == DAYS) {
            return plus(Math.multiplyExact(amountToAdd, SECONDS_PER_DAY), 0);
        }
        if (unit.isDurationEstimated()) {
            throw new UnsupportedTemporalTypeException("Unit must not have an estimated duration");
        }
        if (amountToAdd == 0) {
            return this;
        }
        if (unit instanceof ChronoUnit) {
            switch ((ChronoUnit) unit) {
            case NANOS:
                return plusNanos(amountToAdd);
            case MICROS:
                return plusSeconds((amountToAdd / (1000_000L * 1000)) * 1000)
                        .plusNanos((amountToAdd % (1000_000L * 1000)) * 1000);
            case MILLIS:
                return plusMillis(amountToAdd);
            case SECONDS:
                return plusSeconds(amountToAdd);
            }
            return plusSeconds(Math.multiplyExact(unit.getDuration().seconds, amountToAdd));
        }
        Duration duration = unit.getDuration().multipliedBy(amountToAdd);
        return plusSeconds(duration.getSeconds()).plusNanos(duration.getNano());
    }
```
it shows that DAYS is the case exempted from `isDurationEstimated`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -226,50 +255,76 @@ private void validateAsynchronousAnnotation(MethodDeclaration node, MarkerAnnota
...
+					? (SimpleName) memberUnitExpr
+					: ((QualifiedName) memberUnitExpr).getName();
+			memberUnit = memberUnitName.getIdentifier();
+		}
+		Duration memberValue = null;
+		try {
+			memberValue = memberUnit != null
+					? Duration.of(memberUnitNum, ChronoUnit.valueOf(memberUnit))
+					: Duration.of(memberUnitNum, ChronoUnit.MILLIS);
+		} catch (UnsupportedTemporalTypeException e) {
```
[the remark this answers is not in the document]
AlexXuChen: I believe this would be a syntax error on the java side then.

i.e. consider what you entered:
```java
@Retry(delay = 1, maxDuration = 1, delayUnit = ChronoUnit.HOU)
```
Then the client will have the following error:
```
HOU cannot be resolved or is not a fieldJava(33554502)
```

If this is the case, do we still need to handle illegal arguments?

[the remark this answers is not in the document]
AlexXuChen: If we do handle the illegal arguments, I assume it covers type mismatch, cannot be resolved to variable, etc..

Issue #196, opened 2021-10-06
Support remaining date based ChronoUnit enumerations for `@Retry` annotation


AlexXuChen:
As per https://github.com/eclipse/lsp4mp/pull/193#issuecomment-934591227:

We use `java.time.Duration` to calculate the difference between effective delay and max duration when passed as member values in the `@Retry` annotation. According to the implementation of `Duration.plus`, any date based `ChronoUnit` enumeration excluding `DAYS` (e.g. `WEEKS`, `MONTHS`, `YEARS`, etc..) will throw an `UnsupportedTemporalTypeException`.

Seeing that this is a limitation to the current implementation for `@Retry` diagnostics, (where we simply ignore these cases), there are 2 possible solutions:
1. Have a wrapper to handle every date based case
2. Wait for `java.time.Duration` to support date based enumeration(s)

rgrunber:
Would this work ?

```
import java.time.temporal.ChronoUnit;
public double getDurationInNanos (ChronoUnit unit, long unitValue) {
  double seconds = (double) unit.getDuration().getSeconds();
  int nanos = unit.getDuration().getNano();
  return (seconds * 1000000000 * unitValue) + (nanos * unitValue);
}
```
```
jshell> getDurationInNanos(ChronoUnit.YEARS, 9) > getDurationInNanos(ChronoUnit.DECADES, 10)
$5 ==> false
jshell> getDurationInNanos(ChronoUnit.YEARS, 11) > getDurationInNanos(ChronoUnit.DECADES, 1)
$7 ==> true
```

PR #197, opened 2021-10-06
removed Duration class from `@Retry` annotation and supported date based `ChronoUnit`

AlexXuChen:
Removed `Duration` class from `@Retry` annotation and supported date based `ChronoUnit` enumeration(s)

Fixes: #196 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
@@ -34,9 +34,7 @@ import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFa
 import static org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceErrorCode.FAULT_TOLERANCE_DEFINITION_EXCEPTION;
 
 import java.text.MessageFormat;
-import java.time.Duration;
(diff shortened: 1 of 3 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -276,34 +274,43 @@ private void validateRetryAnnotation(NormalAnnotation annotation)
...
 						DiagnosticSeverity.Warning);
 			}
 		}
 	}
 
-	private Duration findDurationUnit(Expression memberUnitExpr,
+	private int castMemberToInt(Expression expr) {
+		if (expr != null) {
+			try {
+				return (int) expr.resolveConstantExpressionValue();
```
rgrunber, line 304: For me, `expr.resolveConstantExpressionValue()` returns null, and the cast to int is what throws the NPE. To avoid that just get the result of `expr.resolveConstantExpressionValue()` and check whether its null. If it is, for any of delay, maxDuratio or jitter values, then we can avoid emitting a diagnostic since we can't cast the value to an integer. Most likely JDT will handle showing the correct diagnostic.

AlexXuChen: I believe it's okay if jitter is null, delay and maxDuration should be checked for null.

rgrunber: Ok, agreed.

PR #203, opened 2021-10-12
Support for default value inside properties expression

### LINKED ISSUE #201: Support for default value inside properties expression
Parser should support the default value for expression to fix https://github.com/redhat-developer/vscode-quarkus/issues/392

I think xe should ignore unkwown error for property expression which define default value.

angelozerr:
Support for default value inside properties expression

Fixes #201

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyValueExpression.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyValueExpression.java
@@ -24,8 +24,20 @@ import org.eclipse.lsp4mp.utils.PropertiesFileUtils;
  * When properties file is processed, the reference is replaced with the value
  * of the other property. In the properties file, it has the form:
  * <code>${other.property.name}</code>
+ * 
+ * @see https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#property-expressions
  */
 public class PropertyValueExpression extends Node {
+
+	private boolean parsed;
+
+	private int referenceNameStartOffset = -1;
+
+	private int referenceNameEndOffset = -1;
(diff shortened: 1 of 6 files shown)


angelozerr:
At first you can find property expressionspecification at https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#property-expressions 

Please note that we don't support composed property expression. I created an issue for that https://github.com/eclipse/lsp4mp/issues/204

To test this PR, open an application.properties and add:

```
quarkus.application.name=${name}
quarkus.application.version=${quarkus.application.name}
```

you should see an error on name:

image

And if you hover the `quarkus.application.version` you should see the value wich is not resolved:

image

Now if you define a default value to name like this:

```
quarkus.application.name=${name:XXX}
quarkus.application.version=${quarkus.application.name}

```

You should have none errors and if you hover the `quarkus.application.version` you should see the value wich is resolved:

image

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/PropertyValueExpression.java
```diff
@@ -105,38 +118,97 @@ public String getReferencedPropertyName() {
...
+		}
+		if (start >= end) {
+			return;
+		}
+		boolean nameParsing = true;
+		referenceNameStartOffset = start + 2;
+		String text = super.getOwnerModel().getText();
+		for (int i = referenceNameStartOffset; i < end; i++) {
+			char c = text.charAt(i);
+			switch (c) {
```
AlexXuChen, line 189: I believe the offset ensures we start scanning after `${`, then the cases are treated accordingly. If we have a name as follows:
```
quarkus.application.name=${abc:def:XXX}
```
The hover will only recognize after the last :
Screenshot from 2021-10-13 13-57-15

Should we permit this behaviour?

PR #206, opened 2021-10-18
Reference only property declared in *.properties file in property expression

### LINKED ISSUE #205: Reference only property declared in *.properties file in property expresion
After playing with property expression and reading spec at https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#property-expressions it seems that we cannot reference property which comes from @ConfigProperty:

`A default value defined via org.eclipse.microprofile.config.inject.ConfigProperty#defaultValue is not eligible to be expanded since multiple candidates may be available.`

It means that if we have in Java file:

```java
@ConfigProperty(name="DBUSER" defaultValue="sa")
private String dbuser:
```

It's forbidden to reference this property (the Quarkus application start fails):

`my.property = ${DBUSER}`

If we write that:

`
DBUSER = sa
my.property = ${DBUSER}
`

It's allowed.

Today we provide DBUSER and another properties quarkus.*, mp.* etc when we do completion inside a property expression

`my.property = ${ |`

which is wrong. We should fix completion and we should report an error if we write for instance

`my.property = ${quarkus.application.name}`

where quarkus.application.name is a computed property coming from Java.

angelozerr:
Reference only property declared in *.properties file in property expresion

Fixes #205

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
@@ -410,13 +410,16 @@ class PropertiesFileCompletions {
 		for (String independentProperty : independentProperties) {
 			list.getItems().add(getPropertyCompletionItem(independentProperty, (PropertyValueExpression) node, model));
 		}
+
 		// Add all properties not referenced in the properties file as completion
-		// options
+		// options only the property has no default value
 		for (ItemMetadata candidateCompletion : projectInfo.getProperties()) {
-			String candidateCompletionName = candidateCompletion.getName();
-			if (!graph.hasNode(candidateCompletionName)) {
-				list.getItems()
-						.add(getPropertyCompletionItem(candidateCompletionName, (PropertyValueExpression) node, model));
+			if (candidateCompletion.getDefaultValue() == null) {
+				String candidateCompletionName = candidateCompletion.getName();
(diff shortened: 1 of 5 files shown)


radcortez:
> i.e. i'm surprised if programmatically defined properties can't be referred to ...thats news to me (but I kinda understand why ... but just wonder how you refer to http.port in a url without having to override the default ?)

They can. They just need to be valid on runtime, when the expression is evaluated. 

The message `Reference the property [property] declared in Java is not allowed ?` is not coming from Config. So, something else is doing that validation.

maxandersen:
@radcortez yeah so what you are saying (  which is what I thought too ) is that what is described in https://github.com/eclipse/lsp4mp/issues/205 does not hold true in all cases ? (not sure if spec is more restrictive and we allow it or @angelozerr interpreted the spec too strict ?)

angelozerr:
>  (not sure if spec is more restrictive and we allow it or @angelozerr interpreted the spec too strict ?)

@maxandersen according the spec at https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#property-expressions it seems that we cannot reference property which comes from @ConfigProperty:

`A default value defined via org.eclipse.microprofile.config.inject.ConfigProperty#defaultValue is not eligible to be expanded since multiple candidates may be available.`

I tested this usecase in Quarkus application and when I try to reference a property coming from @ConfigProperty, the application cannot be started. It's the reason why I wanted to manage that,because I spend some time to understand why my Quarkus application could not be started. 

@radcortez you mean that it's not a correct validation?

radcortez:
Sounds good :)

PR #211, opened 2021-11-19
Support optional property reference hover for annotation members

AlexXuChen:
Support optional property reference hover for annotation members.

Fixes [#redhat-developer/quarkus-ls/401](https://github.com/redhat-developer/quarkus-ls/issues/401)

Requires https://github.com/redhat-developer/quarkus-ls/pull/431 to test

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
@@ -16,9 +16,8 @@ package org.eclipse.lsp4mp.jdt.core.java.hover;
 import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.getAnnotation;
 import static org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.getAnnotationMemberValue;
 
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -59,7 +58,9 @@ public class PropertiesHoverParticipant implements IJavaHoverParticipant {
 
 	private final String defaultValueAnnotationMemberName;
 
-	private final String [] annotationMembers;
+	private final String[] annotationMembers;
+
+	private Function<String, String> propertyReplacer;
 
 	public PropertiesHoverParticipant(String annotationName, String annotationMemberName) {
 		this(annotationName, annotationMemberName, null);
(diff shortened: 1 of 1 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/hover/PropertiesHoverParticipant.java
```diff
@@ -158,7 +159,7 @@ public Hover collectHover(JavaHoverContext context, IProgressMonitor monitor) th
 		}
 
 		// property references may be surrounded by curly braces in Java file
-		propertyKey = propertyKey.replace("{", "").replace("}", "");
+		propertyKey = propertyKey.replace("{", "").replace("}", "").replace("$", "");
```
rgrunber, line 162: In https://github.com/redhat-developer/quarkus-ls/issues/425 , we discussed that this isn't the right approach because it applies to every annotation property. Certain properties like `cron` support `{..}` and `${..}` but others like `name` from `@ConfigProperty` don't and will produce unwanted behaviour.

@angelozerr has a comment, https://github.com/redhat-developer/quarkus-ls/issues/425#issuecomment-961177573 , that explains how to go about fixing this.

AlexXuChen: Ah gotcha, so separate them

AlexXuChen:
If you hover `@ConfigProperty(name="${greeting.message}")` or  `@ConfigProperty(name="{greeting.message}")`, the display shows:
```
The property '${greeting.message}' is not assigned a value in any config file, and must be assigned at runtime.microprofile-config(NO_VALUE_ASSIGNED_TO_PROPERTY)
${greeting.message} is not set
```
respectively.

Is this the intended behaviour @rgrunber?

rgrunber:
> If you hover `@ConfigProperty(name="${greeting.message}")` or `@ConfigProperty(name="{greeting.message}")`, the display shows:
> 
> ```
> The property '${greeting.message}' is not assigned a value in any config file, and must be assigned at runtime.microprofile-config(NO_VALUE_ASSIGNED_TO_PROPERTY)
> ${greeting.message} is not set
> ```
> 
> respectively.
> 
> Is this the intended behaviour @rgrunber?

Yes, unless you defined a property `${greeting.message}` in the configuration. The name property is meant to store the property itself so there's no need to use the syntax to reference it.

rgrunber (changes requested):
Change works well for me. Could you try adding a testcase in `QuarkusConfigJavaHoverTest` that confirms that `@ConfigProperty(name=${greeting.message})` does not produce a hover ? Not sure if it's possible, but maybe you can test what the result of a hover is when the reference is not defined. You can just add the field in config-hover's GreetingResource.

rgrunber:
Also, I would rename the commit + PR title to something that doesn't mention Quarkus since this is in lsp4mp.

"Support optional property references for annotation members" ?

AlexXuChen:
> Change works well for me. Could you try adding a testcase in `QuarkusConfigJavaHoverTest` that confirms that `@ConfigProperty(name=${greeting.message})` does not produce a hover ? Not sure if it's possible, but maybe you can test what the result of a hover is when the reference is not defined. You can just add the field in config-hover's GreetingResource.

I'll add the test to redhat-developer/quarkus-ls#431

rgrunber (approved):

PR #213, opened 2021-11-25
Do not rebuild list of configuration properties when MicroProfile config sources are updated in the build directory

### LINKED ISSUE #162: Do not rebuild list of configuration properties when MicroProfile config sources are updated in the build directory
Background:
If you have a config source under `src/main/resources`, there will be another copy of the file representing the built application. For instance, in maven, this is usually under `target/classes`. microprofile-jdt uses the property values from this copy of the file in order to generate completion/validation/hover/etc. in Java files.

Currently, there is a listener defined in `MicroProfilePropertiesListenerManager` that detects a change to these files, and sends an event to microprofile-ls. This event has the same `MicroProfilePropertiesScope` as the event which causes microprofile-jdt to recalculate the list of all config properties by going through the Java files. Since they have the same `MicroProfilePropertiesScope`, microprofile-ls cannot distinguish between them, and must do both whenever it receives the event. This is bad, since calculating the list of all config properties from the Java files is computationally expensive.

Set up a new `MicroProfilePropertiesScope` to signify that one of the config sources has changed, and set it up so that this event only causes Java files to revalidate, and doesn't cause microprofile-jdt to recalculate the list of all properties.

angelozerr:
Do not rebuild list of configuration properties when MicroProfile config sources are updated in the build directory

Fixes #162

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManager.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManager.java
@@ -18,6 +18,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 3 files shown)


rgrunber:
Looks good, and works for me. I just noticed that while the new change only triggers the re-loading with `sources` scope, I see 3 separate threads for the same thing, whereas before there was only one.

To reproduce, open just a Java source file, then save (even without editing anything), and then close the file and open a config property file.

angelozerr:
>  I see 3 separate threads for the same thing, whereas before there was only one. Is this expected ?

No -(

> whereas before there was only one

are you sure?

rgrunber (approved):
Given that this will improve search performance, I'm fine with merging, but the extra search threads seem like they didn't occur before.

angelozerr:
Ok let's merge the PR and see the problem with several threads in an another PR.

PR #224, opened 2022-03-28
Validate default value expression.

### LINKED ISSUE #225: Property file with property expressions are flagged as wrong
### Issue
Property definition is flagged when is assigned an ENV reference with a default value.

It fails with `quarkus.log.level`

image

This appears with Quick Fix:
image

### Extensions version
- VSCode quarkus extension version: 1.10.0
image

- Tools for MicroProfile version: 0.4.0
image

### Related Issues:

- https://github.com/redhat-developer/vscode-quarkus/issues/392
- https://github.com/eclipse/lsp4mp/issues/201

angelozerr:
Validate default value expression

Fixes #225


angelozerr:
Here a demo with the default value support (validation is supported but too completion, hover and definition):

DefaultValueSupport

AlexXuChen (commented):
Seems to be working as described, one potential fix is if we could improve the suggestion(s) for when the quick fix displays for an invalid enum that contains a valid enum as a substring.

i.e. `${ENV:ALLa}` should only suggest `ALL`
Peek 2022-04-04 10-22

angelozerr:
> i.e. ${ENV:ALLa} should only suggest ALL

We have already a did mean code action which does that, it's th ereason we have an another code action which provides all allowed values.

angelozerr:
Thanks @AlexXuChen for your review!

PR #245, opened 2022-05-30
Added JDTTypeUtils helper to check for `void` return type

AlexXuChen:
Added JDTTypeUtils helper to check for void return type.

Fix for https://github.com/redhat-developer/quarkus-ls/pull/636#discussion_r884560801


AlexXuChen:
Fix is in quarkus-ls

PR #247, opened 2022-06-02
Ignored ENV reference in expression validation

### LINKED ISSUE #227: Property file with property expressions (without default value) are flagged as wrong
Continuation of https://github.com/eclipse/lsp4mp/issues/225

### Issue
Property definition is flagged when is assigned an ENV reference without a default value.

image

### Extensions version
- VSCode quarkus extension version: 1.10.0
image

- Tools for MicroProfile version: 0.4.0
image

### Suggestion

Consider omitting the warning/error if:

- Property is defined at `quarkus.kubernetes.env.mapping`
   image
- Property is found locally (System.getenv in java)

 _See https://github.com/eclipse/lsp4mp/issues/225#issuecomment-1086155352_

### Related Issues:
- https://github.com/redhat-developer/vscode-quarkus/issues/392
- https://github.com/eclipse/lsp4mp/issues/201
- https://github.com/eclipse/lsp4mp/issues/225

AlexXuChen:
Ignored ENV reference for expression validation without the addition of a default value.
Screenshot from 2022-06-02 10-19-21

Fixes #227 

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
@@ -38,6 +38,7 @@ import org.eclipse.lsp4mp.model.PropertiesModel;
 import org.eclipse.lsp4mp.model.Property;
 import org.eclipse.lsp4mp.model.PropertyValueExpression;
 import org.eclipse.lsp4mp.settings.MicroProfileValidationSettings;
+import org.eclipse.lsp4mp.utils.EnvUtils;
 import org.eclipse.lsp4mp.utils.PositionUtils;
 import org.eclipse.lsp4mp.utils.PropertiesFileUtils;
 
@@ -240,20 +241,21 @@ class PropertiesFileValidator {
 							}
 						} else {
 							if (propValExpr.hasDefaultValue()) {
+								// The expression has default value (ex : ${DBUSER:sa})
 								int start = propValExpr.getDefaultValueStartOffset();
 								int end = propValExpr.getDefaultValueEndOffset();
 								validatePropertyValue(propertyName, metadata, propValExpr.getDefaultValue(), start, end,
 									propValExpr.getOwnerModel());
 							} else {
-								// The expression has default value (ex : ${DBUSER:sa}) otherwise the error
-								// is reported
-								Range range = PositionUtils.createRange(propValExpr.getReferenceStartOffset(),
-									propValExpr.getReferenceEndOffset(), propValExpr.getDocument());
-								if (range != null) {
-									addDiagnostic("Unknown referenced property value expression '" + refdProp + "'",
-										range, expressionSeverity, ValidationType.expression.name());
-								} else {
-
+								if (!(EnvUtils.isEnvVariable(refdProp))) {
+									// or the expression is an ENV variable
+									// otherwise the error is reported
+									Range range = PositionUtils.createRange(propValExpr.getReferenceStartOffset(),
+										propValExpr.getReferenceEndOffset(), propValExpr.getDocument());
+									if (range != null) {
+										addDiagnostic("Unknown referenced property value expression '" + refdProp + "'",
+											range, expressionSeverity, ValidationType.expression.name());
+									}
 								}
 							}
 						}
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/EnvUtils.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/EnvUtils.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.utils;
+
+/**
+ * Environment variable utility class
+ *
+ */
+public class EnvUtils {
+	/**
+	 * Return true if the variable is an ENV variable, and false otherwise
+	 *
+	 * @param variable the variable to check
+	 * @return true if the variable is an ENV variable, and false otherwise
+	 */
+	public static boolean isEnvVariable(String variable) {
+		char[] chars = variable.toCharArray();
+		for (char c : chars) {
+			if (!Character.isUpperCase(c) && Character.isLetter(c)) {
+				return false;
+			}
+		}
+		return true;
+	}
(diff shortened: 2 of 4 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
```diff
@@ -240,20 +240,21 @@ private void validatePropertyValueExpressions(String propertyName, ItemMetadata
...
-								// The expression has default value (ex : ${DBUSER:sa}) otherwise the error
-								// is reported
-								Range range = PositionUtils.createRange(propValExpr.getReferenceStartOffset(),
-									propValExpr.getReferenceEndOffset(), propValExpr.getDocument());
-								if (range != null) {
-									addDiagnostic("Unknown referenced property value expression '" + refdProp + "'",
-										range, expressionSeverity, ValidationType.expression.name());
-								} else {
-
+								if (!(Pattern.matches("[A-Z_]+[1-9A-Z_]*", refdProp.toString()))) {
```
angelozerr, line 249: Try to avoid using Pattern which can be slow. If you need really using pattern , compile it at first. But in your case you don't need to use regexp. You can loop for characters of refProp and call `Character.isUpperCase`.

I suggest that you create a private static method  `isEnvVariable(String variable)` which does that.

AlexXuChen: I used a pattern since `isUpperCase` simply checks character case, and misses validating special characters in ENVs. i.e. an ENV variable can include numbers and _, but cannot begin with numbers. Do you want me to make the validation less strict?

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
```diff
@@ -240,20 +240,21 @@ private void validatePropertyValueExpressions(String propertyName, ItemMetadata
...
-								// The expression has default value (ex : ${DBUSER:sa}) otherwise the error
-								// is reported
-								Range range = PositionUtils.createRange(propValExpr.getReferenceStartOffset(),
-									propValExpr.getReferenceEndOffset(), propValExpr.getDocument());
-								if (range != null) {
-									addDiagnostic("Unknown referenced property value expression '" + refdProp + "'",
-										range, expressionSeverity, ValidationType.expression.name());
-								} else {
-
+								if (!(isEnvVariable(refdProp.toString()))) {
```
angelozerr, line 249: No need to use toString

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
```diff
@@ -419,6 +420,16 @@ private static boolean isBuildtimePlaceholder(String str) {
...
 	}
 
+	/**
+	 * Return true if the variable is an ENV variable, and false otherwise
+	 *
+	 * @param variable the variable to check
+	 * @return true if the variable is an ENV variable, and false otherwise
+	 */
+	private static boolean isEnvVariable(String variable) {
+		return variable.toUpperCase().equals(variable);
```
angelozerr, line 430: variable.toUpperCase() will create a new instance of String (memory problem).

Why you could not do write like this:

```java
public static boolean isEnvVariable(String variable) {
	char[] chars = variable.toCharArray();
	for (char c : chars) {
		if (Character.isLetter(c)) {
			return Character.isUpperCase(c);
		}
	}
	return true;
}
```

I think it should be better to move this method in a new class like` com.redhat.qute.utils.EnvUtils` and write several test with this method. It i simportant to have those kind of test because if we change the implementation or the logic we will be sure that we don't break anything.

AlexXuChen: This would be in lsp4mp correct?

angelozerr: Ooops sorry, I wanted to say `org.eclipse.lsp4mp.utils.EnvUtils`

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/EnvUtils.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public class EnvUtils {
+	/**
+	 * Return true if the variable is an ENV variable, and false otherwise
+	 *
+	 * @param variable the variable to check
+	 * @return true if the variable is an ENV variable, and false otherwise
+	 */
+	public static boolean isEnvVariable(String variable) {
+		return variable.toUpperCase().equals(variable);
```
angelozerr, line 28: Why you don't use Characher.isUpperCase like I suggested in https://github.com/eclipse/lsp4mp/pull/247#discussion_r888985405 

`variable.toUpperCase()` will create a new instance of String, it is real shame to do that just to check that variable is uppercase.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/EnvUtils.java
```diff
@@ -0,0 +1,36 @@
...
+	 * Return true if the variable is an ENV variable, and false otherwise
+	 *
+	 * @param variable the variable to check
+	 * @return true if the variable is an ENV variable, and false otherwise
+	 */
+	public static boolean isEnvVariable(String variable) {
+		char[] chars = variable.toCharArray();
+		for (char c : chars) {
+			if (Character.isLetter(c)) {
+				return Character.isUpperCase(c);
```
rgrunber, line 31: So then `Somevalue` is an environment variable according to this logic, because we decide on the first value only.

You can determine something is not an environment variable if some character fails the test, and "short circuit" the later checks, but to confirm it is an environment variable, you would need to check every letter-based character.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/EnvUtils.java
```diff
@@ -0,0 +1,36 @@
...
+	/**
+	 * Return true if the variable is an ENV variable, and false otherwise
+	 *
+	 * @param variable the variable to check
+	 * @return true if the variable is an ENV variable, and false otherwise
+	 */
+	public static boolean isEnvVariable(String variable) {
+		char[] chars = variable.toCharArray();
+		for (char c : chars) {
+			if (!Character.isUpperCase(c) && !Character.isDigit(c) && c != '_') {
```
rgrunber, line 30: Would `!Character.isUpperCase(c) && Character.isLetter(c)` work ?

AlexXuChen: It will work, just allows for special characters other than `_`

rgrunber (approved):
Works for me. capitalized environment variables are left alone, as expected. There's quite a few strange things that can be written as a variable, but as discussed, that is better left to some future validation.

PR #248, opened 2022-06-08
Temporarily use JDT-LS 1.12.0 (instead of SNAPSHOT build)

rgrunber:
- Eventually a proper update to Java 17 will be required


angelozerr:
Thanks @rgrunber

PR #253, opened 2022-06-13
corrected path to javaASTValidator in schema

### LINKED ISSUE #234: javaASTValidator schema refers to the wrong interface name
https://github.com/eclipse/lsp4mp/blob/f54b7355be2797ec8843e7d106cb4af0296a2ba6/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/schema/javaASTValidators.exsd#L64

JessicaJHee:
Fixes #234

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/schema/javaASTValidators.exsd
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/schema/javaASTValidators.exsd
@@ -61,7 +61,7 @@
                   The Java class for AST Validator.
                </documentation>
                <appinfo>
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
Looks good. Just one thing to mention. Your commit title is :

`corrected path to javaASTValidator in schema`

It's usually good practice to captialize the first word, and write in the imperative (eg. Fix bug vs. Fixed bug). In this case, because "Correct path to .." might be a bit confusing, I would use "Fix", or "Adjust".

These aren't really in the contributor guide, but they're rules that would serve you well on pretty much any project. See https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html for some more details.

angelozerr:
Thanks @JessicaJHee !

PR #256, opened 2022-06-16
Update lsp4j 0.14.0

### LINKED ISSUE #254: Move to LSP4j 0.14.0
Like we did for LemMinx with LSP4J 0.13.0 https://github.com/eclipse/lemminx/pull/1205/files

It seems now last version of LSP4J is 0.14, so we need to change it for Quarks LS and Lemminx too.

datho7561:
Closes #254


rgrunber (approved):
I think this change should be fine.

As a side note, the extensions to JDT-LS in `lsp4mp/microprofile.jdt` or `quarkus-ls/{quarkus,qute}.jdt.ext` use the JDT-LS target platform, so they will continue to use an older lsp4j : https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target#L38

PR #258, opened 2022-06-17
Add check to ensure prefix used is from ConfigProperties

### LINKED ISSUE #233: MicroProfileConfigASTValidator may use prefix from annotation which is not @ConfigProperties
https://github.com/eclipse/lsp4mp/blob/f54b7355be2797ec8843e7d106cb4af0296a2ba6/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java#L109

JessicaJHee:
Fixes #233 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.lsp4mp.jdt.internal.config.java;
 
+import static org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants.CONFIG_PROPERTIES_ANNOTATION;
 import static org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants.CONFIG_PROPERTIES_ANNOTATION_PREFIX;
 import static org.eclipse.lsp4mp.jdt.core.MicroProfileConfigConstants.CONFIG_PROPERTY_ANNOTATION;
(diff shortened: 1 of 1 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
```diff
@@ -105,15 +107,17 @@ public boolean visit(TypeDeclaration typeDeclaration) {
 		List modifiers = typeDeclaration.modifiers();
 		for (Object modifier : modifiers) {
 			if (modifier instanceof NormalAnnotation) {
-				try {
-					Expression prefixExpr = AnnotationUtils.getAnnotationMemberValueExpression(
-							(NormalAnnotation) modifier, CONFIG_PROPERTIES_ANNOTATION_PREFIX);
-					if (prefixExpr != null) {
-						currentPrefix = ((StringLiteral) prefixExpr).getLiteralValue();
+				if (AnnotationUtils.isMatchAnnotation((IAnnotation) modifier, CONFIG_PROPERTIES_ANNOTATION_PREFIX)) {
```
angelozerr, line 110: I think yiu should use CONFIG_PROPERTIES_ANNOTATION no?

rgrunber: Also spoke to @JessicaJHee about casting :)

rgrunber (approved):
LGTM. Note that it could have also been possible to write a testcase for this but I think we can accept this. ie. use an annotation that is not `@ConfigProperties` but declares a `prefix = "foo"` and assert that no diagnostic is generated for a given field that is `@ConfigProperty` annotated with no default value (or config file definition) that states the property is prefixed with `foo`.

PR #260, opened 2022-06-22
Add checks to prevent ClassCastException for incorrect retry value

### LINKED ISSUE #235: Invalid value in Retry annotation may lead to ClassCastException
This code is making the assumption that user entered an integer value but if he entered something else that would lead to ClassCastException

https://github.com/eclipse/lsp4mp/blob/f54b7355be2797ec8843e7d106cb4af0296a2ba6/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java#L287

JessicaJHee:
Fixes #235 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
@@ -57,6 +57,7 @@ import org.eclipse.jdt.core.dom.Expression;
 import org.eclipse.jdt.core.dom.MarkerAnnotation;
 import org.eclipse.jdt.core.dom.MethodDeclaration;
 import org.eclipse.jdt.core.dom.NormalAnnotation;
+import org.eclipse.jdt.core.dom.NumberLiteral;
 import org.eclipse.jdt.core.dom.QualifiedName;
 import org.eclipse.jdt.core.dom.SimpleName;
 import org.eclipse.jdt.core.dom.Type;
@@ -274,21 +275,19 @@ public class MicroProfileFaultToleranceASTValidator extends JavaASTValidator {
 			Expression jitterUnitExpr = getAnnotationMemberValueExpression(
 					annotation, JITTER_DELAY_UNIT_RETRY_ANNOTATION_MEMBER);
 
-			Object delayConstant = delayExpr != null
-					? delayExpr.resolveConstantExpressionValue()
+			Object delayConstant = delayExpr != null ? delayExpr.resolveConstantExpressionValue() : null;
+			Object maxDurationConstant = maxDurationExpr != null ? maxDurationExpr.resolveConstantExpressionValue()
 					: null;
-			Object maxDurationConstant = maxDurationExpr != null
-					? maxDurationExpr.resolveConstantExpressionValue()
-					: null;
-			Object jitterConstant = jitterExpr != null
-					? jitterExpr.resolveConstantExpressionValue()
-					: null;
+			Object jitterConstant = jitterExpr != null ? jitterExpr.resolveConstantExpressionValue() : null;
 
-			int delayNum = delayConstant != null ? (int) delayConstant : -1;
-			int maxDurationNum = maxDurationConstant != null
-					? (int) maxDurationConstant
-					: -1;
-			int jitterNum = jitterConstant != null ? (int) jitterConstant : 0;
+			long delayNum = delayConstant instanceof Integer ? (long) (int) delayConstant
+					: (delayConstant instanceof Long ? (long) delayConstant : -1);
+
+			long maxDurationNum = maxDurationConstant instanceof Integer ? (long) (int) maxDurationConstant
+					: (maxDurationConstant instanceof Long ? (long) maxDurationConstant : -1);
+
+			long jitterNum = jitterConstant instanceof Integer ? (long) (int) jitterConstant
+					: (jitterConstant instanceof Long ? (long) jitterConstant : 0);
 
 			if (delayNum != -1 && maxDurationNum != -1) {
 				double delayValue = findDurationUnit(delayUnitExpr, delayNum);
@@ -310,7 +309,7 @@ public class MicroProfileFaultToleranceASTValidator extends JavaASTValidator {
 	}
 
 	private double findDurationUnit(Expression memberUnitExpr,
-			int memberUnitNum) {
+			long memberUnitNum) {
 		String memberUnit = null;
 		if (memberUnitExpr != null) {
 			SimpleName memberUnitName = memberUnitExpr instanceof SimpleName
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/projects/maven/microprofile-fault-tolerance/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/invalidParameters/RetryClientForValidation.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/projects/maven/microprofile-fault-tolerance/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/invalidParameters/RetryClientForValidation.java
@@ -85,4 +85,19 @@ public class RetryClientForValidation {
     public Connection validE() {
         return null;
     }
+    
+    @Retry(delay = 1000, maxDuration = 1000, jitter = "invalidType")
+    public Connection invalidG() {
+        return null;
+    }
+    
+    @Retry(delay = 1000, maxDuration = "invalidType")
+    public Connection invalidH() {
+        return null;
+    }
+    
+    @Retry(delay = 50000000000L, maxDuration = 1000, jitter = "invalidType")
+    public Connection invalidI() {
+        return null;
+    }
 }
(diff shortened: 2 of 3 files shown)


rgrunber:
The change works for me. All that's left is to write a test for this.

Have a look at `MicroProfileFaultToleranceJavaDiagnosticsTest` and one of the test classes used, like [`RetryClientForValidation.java`](https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/projects/maven/microprofile-fault-tolerance/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/invalidParameters/RetryClientForValidation.java#L54)

When the delay is greater than the maximum duration, we should emit a warning. When one of the values (eg. `jitter`) is a String, the warning wouldn't be emitted since the logic would fail. So maybe adding a case that confirms the warning exists would be good.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -284,11 +284,21 @@ private void validateRetryAnnotation(NormalAnnotation annotation)
...
-			int delayNum = delayConstant != null ? (int) delayConstant : -1;
-			int maxDurationNum = maxDurationConstant != null
-					? (int) maxDurationConstant
-					: -1;
-			int jitterNum = jitterConstant != null ? (int) jitterConstant : 0;
+			int delayNum = -1;
+			int maxDurationNum = -1;
+			int jitterNum = 0;
+
+			if (delayConstant instanceof Integer) {
```
AlexXuChen, line 291: Similar to the ternary assignment before, you can add the instance check as a condition:
```java
int delayNum = (delayConstant != null && delayConstant instanceof Integer) : (int) delayConstant : -1
```
A little less readable, but shorter

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -284,11 +284,21 @@ private void validateRetryAnnotation(NormalAnnotation annotation)
 					? jitterExpr.resolveConstantExpressionValue()
 					: null;
 
-			int delayNum = delayConstant != null ? (int) delayConstant : -1;
-			int maxDurationNum = maxDurationConstant != null
-					? (int) maxDurationConstant
-					: -1;
-			int jitterNum = jitterConstant != null ? (int) jitterConstant : 0;
+			int delayNum = -1;
```
AlexXuChen, line 287: Something I missed initially is the type specification for [Retry](https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.0.2/apidocs/org/eclipse/microprofile/faulttolerance/Retry.html) annotation members. I.e. delay is of type `long`, but we assign it as `int` here. In that case, your case would be using the `Long` wrapper instead. We can change this for consistency. Thoughts @rgrunber?

AlexXuChen (changes requested):
Working as intended, there is no longer a `ClassCastExeption`. Just a note that `Retry` annotation validation is tested here:
https://github.com/eclipse/lsp4mp/blob/63da0d68e7c498ab126c04782faf9bf3a8b43f2d/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/faulttolerance/java/MicroProfileFaultToleranceJavaDiagnosticsTest.java#L253
There might be some value in testing what happens when the user inputs the incorrect type.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
rgrunber: I just tried creating a number that's larger than the int range, and when I do JDT complains `The literal 5000000000 of type int is out of range Java(536871066)` . So this might be a bug in JDT. I think it should be fine for now.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
rgrunber: Makes sense to do. I think you mean `?` for the first `:`, but otherwise I think this should be nicer.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
rgrunber: After reproducing this, I think I know what JDT expects. Instead of providing a number larger than an `int` (eg. a `long`) as just "the number", it expects it in the form `5000000000L`. Then it won't complain. I guess it needs the user to explicitly mark it as long.

Once that is done, `resolveConstantExpressionValue(..)` will return a `Long`, not an `Integer` so we could either :
- Handle both `Long` and `Integer` as separate cases
- See if you can combine all cases using `Long.valueOf(...)`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -274,21 +275,19 @@ private void validateRetryAnnotation(NormalAnnotation annotation)
...
+			Object maxDurationConstant = maxDurationExpr != null ? maxDurationExpr.resolveConstantExpressionValue()
 					: null;
+			Object jitterConstant = jitterExpr != null ? jitterExpr.resolveConstantExpressionValue() : null;
 
-			int delayNum = delayConstant != null ? (int) delayConstant : -1;
-			int maxDurationNum = maxDurationConstant != null
-					? (int) maxDurationConstant
-					: -1;
-			int jitterNum = jitterConstant != null ? (int) jitterConstant : 0;
+			long delayNum = delayConstant instanceof Integer ? (long) (int) delayConstant
```
AlexXuChen, line 283: Casting to an int, then a long after determining that the type is an integer is the equivalent to just casting to a long. Thus the `instanceof Long` check will always pass. To handle both cases separately, it might be better to check each condition separately then using a ternary assignment.

AlexXuChen (changes requested):

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
rgrunber, line 283: This all works. Just one thing I noticed to maybe save a bit of space :

```
jshell> Integer int_val = 3;
int_val ==> 3
jshell> long res = (long) (int) int_val
res ==> 3
jshell> long res = (long) int_val
res ==> 3

jshell> Long long_val = 3L;
long_val ==> 3
jshell> long res = long_val;
res ==> 3
```

I guess the interesting part is that auto-boxing need not happen between the same types.

rgrunber: Oops, so if it does work, you can probably simplify this a lot, and just assign the value if `Integer` or `Long`.

rgrunber: The fact that we're coming form `Object` makes a difference. I guess we shouldn't try to play too much with the auto-boxing feature.

```
jshell> Object delay1 = 3;
delay1 ==> 3

jshell> Integer delay2 = 4;
delay2 ==> 4

jshell> (long) delay1
|  Exception java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
|        at (#12:1)

jshell> (int) delay1
$13 ==> 3

jshell> (long) delay2
$14 ==> 4

jshell> (int) delay2
$15 ==> 4
```

rgrunber (approved):

PR #261, opened 2022-06-22
Add check to prevent failure on incorrect attribute values

### LINKED ISSUE #249: AnnotationAtttributeRule may fail on attribute values
https://github.com/eclipse/lsp4mp/blob/2d7c54fb9c04d640bc778d51bac0300be970fb33/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationAttributeRule.java#L87

This class is parsing the attribute value as a double. But the value maybe 123L or 2 * 3 or 1_000_000 which would cause an exception.

See https://github.com/redhat-developer/intellij-quarkus/issues/596

JessicaJHee:
Fixes #249 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
@@ -22,7 +22,10 @@ import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.dom.Annotation;
 import org.eclipse.jdt.core.dom.Expression;
+import org.eclipse.jdt.core.dom.InfixExpression;
 import org.eclipse.jdt.core.dom.NormalAnnotation;
+import org.eclipse.jdt.core.dom.NumberLiteral;
+import org.eclipse.jdt.core.dom.PrefixExpression;
 import org.eclipse.jdt.core.dom.SingleMemberAnnotation;
 import org.eclipse.lsp4j.DiagnosticSeverity;
 import org.eclipse.lsp4mp.jdt.core.java.validators.JavaASTValidator;
@@ -96,12 +99,17 @@ public class AnnotationRulesJavaASTValidator extends JavaASTValidator {
 		if (attributeValueExpr == null) {
 			return;
 		}
+		// Ensure value of AST attribute is a valid integer or an expression that can be evaluated
+		if (!isInteger(attributeValueExpr) && !isInfixIntegerExpression(attributeValueExpr)) {
+			return;
+		}
 
-		// Get the value of the AST attribute
-		String valueAsString = attributeValueExpr.toString();
+		// Resolve the value of the AST attribute
+		String valueAsString = attributeValueExpr.resolveConstantExpressionValue().toString();
 		if (StringUtils.isEmpty(valueAsString)) {
 			return;
 		}
+
 		// Validate the value with the rule
 		String validationResult = JavaASTValidatorRegistry.getInstance().validate(valueAsString, attributeRule);
 		if (validationResult != null) {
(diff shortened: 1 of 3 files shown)


JessicaJHee:
I'm not sure if any tests are required for this?

rgrunber:
Yeah, I don't think it's possible to test this. I would maybe add a comment for that if statement to explain what each condition is testing. eg. a number (int or long), a negative number.

datho7561 (approved):
From my testing, this works. I don't get any errors in the logs when I use expressions for the fields of the Retry annotation. Well done!

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
```diff
@@ -99,9 +102,17 @@ private void validateAnnotationAttributeValue(Expression attributeValueExpr, Ann
 
 		// Get the value of the AST attribute
 		String valueAsString = attributeValueExpr.toString();
-		if (StringUtils.isEmpty(valueAsString)) {
+
+		// Ensure value of AST attribute is a valid integer or an expression that can be evaluated
+		if ((StringUtils.isEmpty(valueAsString)) || (!(attributeValueExpr instanceof NumberLiteral)
+				&& !(attributeValueExpr instanceof PrefixExpression && isSignedInteger(attributeValueExpr))
+				&& !(attributeValueExpr instanceof InfixExpression && isInfixIntegerExpression(attributeValueExpr)))) {
```
rgrunber, line 109: Could you also move the `attributeValueExpr instanceof ...` within the corresponding helper method ? Given that the methods take just an `Expression`, it should work and keep things even simpler.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
```diff
@@ -99,9 +102,16 @@ private void validateAnnotationAttributeValue(Expression attributeValueExpr, Ann
...
+
+		// Ensure value of AST attribute is a valid integer or an expression that can be evaluated
+		if (StringUtils.isEmpty(valueAsString)
+				|| (!isInteger(attributeValueExpr) && !isInfixIntegerExpression(attributeValueExpr))) {
 			return;
 		}
+
+		if (attributeValueExpr instanceof InfixExpression) {
+			valueAsString = attributeValueExpr.resolveConstantExpressionValue().toString();
+		}
```
rgrunber, line 114: I would make the following change :

```diff
diff --git a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
index 84d823c..df13bef 100644
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/annotations/AnnotationRulesJavaASTValidator.java
@@ -100,18 +100,18 @@ public class AnnotationRulesJavaASTValidator extends JavaASTValidator {
 			return;
 		}
 
-		// Get the value of the AST attribute
-		String valueAsString = attributeValueExpr.toString();
-
 		// Ensure value of AST attribute is a valid integer or an expression that can be evaluated
-		if (StringUtils.isEmpty(valueAsString)
-				|| (!isInteger(attributeValueExpr) && !isInfixIntegerExpression(attributeValueExpr))) {
+		if (!isInteger(attributeValueExpr) && !isInfixIntegerExpression(attributeValueExpr)) {
 			return;
 		}
 
-		if (attributeValueExpr instanceof InfixExpression) {
-			valueAsString = attributeValueExpr.resolveConstantExpressionValue().toString();
+		// Get the value of the AST attribute
+		String valueAsString = attributeValueExpr.resolveConstantExpressionValue().toString();
+
+		if (StringUtils.isEmpty(valueAsString)) {
+			return;
 		}
+
 		// Validate the value with the rule
 		String validationResult = JavaASTValidatorRegistry.getInstance().validate(valueAsString, attributeRule);
 		if (validationResult != null) {
```

`resolveConstantExpressionValue()` ca be called even when the expression isn't an infix expression. We can place the check for `isEmpty` after that. Ultimately it's better to avoid calling `toString()` if we can.

rgrunber (approved):
I think the change is ready to be merged after addressing this last point.

PR #262, opened 2022-06-27
Update ASTValidator to use project classpath

### LINKED ISSUE #232: ASTValidator uses IDE classpath instead of project classpath
See https://github.com/eclipse/lsp4mp/blob/f54b7355be2797ec8843e7d106cb4af0296a2ba6/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java#L248

JessicaJHee:
Fixes #232 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
@@ -161,10 +161,11 @@ public class MicroProfileConfigASTValidator extends JavaASTValidator {
 	 */
 	private void validatePropertyDefaultValue(Annotation annotation, Expression defaultValueExpr) {
 		FieldDeclaration fieldDeclaration = (FieldDeclaration) annotation.getParent();
+		IJavaProject javaProject = getContext().getJavaProject();
 		if (defaultValueExpr != null) {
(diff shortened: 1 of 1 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
```diff
@@ -80,12 +80,14 @@ public class MicroProfileConfigASTValidator extends JavaASTValidator {
 	private List<String> patterns;
 	// prefix from @ConfigProperties(prefix="")
 	private String currentPrefix;
+	private static IJavaProject javaProject;
```
rgrunber, line 83: Although this works, there are a few better ways to fix this while also working around some of the restrictions you probably ran into.

- I'd avoid defining `javaProject` because it's coming from `context` and we already have access to it here.
- You could either : 

**(a)** switch `isAssignable` to be an instance method (non-static) and then you'd be able to use `context`
OR
**(b)** add `IJavaProject` as a parameter to `isAssignable` which would keep it as static. You would then have to update any references to ensure they pass in the new parameter.

I like **(b)** because `isAssignable`, can continue to be used in a static sense and doesn't depend on any aspect of the class itself.

angelozerr: You don't need to store the javaProject, you can access it with `super.getContext().getJavaProject();`

PR #263, opened 2022-06-28
Support completion before '='

### LINKED ISSUE #255: Empty completion when completion is triggered before the assign `=`
Given this microprofile-config.properties:

`%dev.=`

If you trigger the completion after `.`

`%dev.|=`

Completion is empty although it should be filled with the all available propeties.

There is the same problem with

`|=`

or 

`quarkus.|=`

or 

`quark|=`

JessicaJHee:
completionBeforeAssign

Fixes #255


JessicaJHee:
I'm not sure how to deal with this test, it seems like it doesn't want any completion results in this case? (expected completion count is 0)

https://github.com/eclipse/lsp4mp/blob/9789a1a996811fade43029605c014c7825e8f1da/microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletionTest.java#L356

rgrunber:
> I'm not sure how to deal with this test, it seems like it doesn't want any completion results in this case? (expected completion count is 0)
> 
> https://github.com/eclipse/lsp4mp/blob/9789a1a996811fade43029605c014c7825e8f1da/microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletionTest.java#L356

The test was probably written to capture the existing behaviour. Since the behaviour will change, it should be updated to reflect that.

JessicaJHee:
> The test was probably written to capture the existing behaviour. Since the behaviour will change, it should be updated to reflect that.

Okay I wasn't sure what the original intent of the test was, I will update it as long the new change doesn't conflict with any expected behavior from before.

angelozerr:
@JessicaJHee please write tests in https://github.com/eclipse/lsp4mp/blob/master/microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletionTest.java

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletionTest.java
```diff
@@ -355,7 +355,13 @@ public void completionDefaultValueContainsDollarSignAndBraces() throws BadLocati
 	@Test
 	public void completionBetweenPropertyNameAndEquals() throws BadLocationException {
 		String value = "quarkus.http.cors|=";
-		testCompletionFor(value, true, 0);
+		testCompletionFor(value, true, 1489);
```
angelozerr, line 358: Please don't hard code the number of expected completion items, because if we add another properties, your test will fail. In our test we use `0` to test that we have none completion. But in your case you expect that you have some completion items. You don't need to check the number of completion items, but just check you have some completion items.

I suggest that you change with (by adjusting the range r(0, 0, 20)):

```
testCompletionFor(value, false,
				c("quarkus.log.category.{*}.level", "quarkus.log.category.{*}.level=inherit", r(0, 0, 20)));
```

I have a question, when you have that 

`greeting.messag|=abcd`

when you apply completion, do you loose the content after `|` ? I mean do you have that:

`greeting.message`

or 

`greeting.message=abcd` ?

You should have the last.

JessicaJHee: It will replace the message after according to the default value. 

completionBeforeAssign

angelozerr: Indeed it was my fear, I think we should not update the value (random_name) when completion is applied.

angelozerr: Ok let's do that in a separate issue.

angelozerr:
Thanks @JessicaJHee !

PR #269, opened 2022-07-05
Remove the m2e lifecycle mapping plugin from MANIFEST.MF

datho7561:
The same functionality is now provided as a part of `org.eclipse.m2e.jdt`.

See https://github.com/eclipse-m2e/m2e-core/commit/8ed00a53d6dba5fb31e81765b73fbeea769fead4


rgrunber:
As a quick explanation of why in the world this was broken for so long :

`org.eclipse.m2e.lifecyclemapping.defaults` was moved into `org.eclipse.m2e.core`. It had its own requirement on `org.eclipse.m2e.jdt` . Now that it's gone, and since we clearly need lifecycle mappings for things like maven-compiler-plugin, we should explicitly depend on it.

PR #270, opened 2022-07-07
Fix application properties completion with existing property value

### LINKED ISSUE #264: Completing property name with existing value will replace current value with default value
Peek 2022-06-30 11-03

Completing the property name should not change the existing value of the property.

JessicaJHee:
completionBeforeAssign

FIxes #264 

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
@@ -186,6 +186,13 @@ class PropertiesFileCompletions {
 
 		Set<String> existingProperties = getExistingProperties(model);
 
+		String propertyValue = null;
+		if (node != null && node.getNodeType() == NodeType.PROPERTY_KEY) {
+			propertyValue = ((PropertyKey) node).getProperty().getPropertyValue();
+		} else if (node != null && node.getNodeType() == NodeType.ASSIGN) {
+			propertyValue = ((Assign) node).getProperty().getPropertyValue();
+		}
+
 		// Completion on MicroProfile properties
 		for (ItemMetadata property : projectInfo.getProperties()) {
 			if (property == null) {
@@ -206,7 +213,13 @@ class PropertiesFileCompletions {
 			CompletionItem item = new CompletionItem(name);
 			item.setKind(CompletionItemKind.Property);
 
-			String defaultValue = property.getDefaultValue();
+			String defaultValue = null;
+			if (propertyValue == null || propertyValue.isEmpty()) {
+				defaultValue = property.getDefaultValue();
+			} else {
+				defaultValue = propertyValue;
+			}
+
 			Collection<ValueHint> enums = PropertiesFileUtils.getEnums(property, projectInfo);
 
 			StringBuilder insertText = new StringBuilder();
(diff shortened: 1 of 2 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
```diff
@@ -229,6 +229,19 @@ private static void collectPropertyKeySuggestions(int offset, Node node, Propert
 				insertText.append(' ');
 			}
 
+			// replace with current property value if exist
+			if (node != null && node.getNodeType() == NodeType.PROPERTY_KEY) {
+				String propertyValue = ((PropertyKey) node).getProperty().getPropertyValue();
+				if (propertyValue != null && propertyValue.isEmpty() == false) {
```
rgrunber, line 235: for a method that returns a boolean, it's common to use the `!` operator to check for `false` (ie. the negation is true).

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
```diff
@@ -229,6 +229,19 @@ private static void collectPropertyKeySuggestions(int offset, Node node, Propert
 				insertText.append(' ');
 			}
 
+			// replace with current property value if exist
```
rgrunber, line 232: I guess this question is directed more at @angelozerr , but if `Assign` inherited from `BasePropertyValue`, then the 2 cases could be combined. Is there a reason it doesn't ? I guess an assignment node need not have a value ?

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
rgrunber: Also, I'm wondering what if the property value is in fact empty. Should we not preserve it anyways ? It might be intentionally empty.

microprofile.ls/org.eclipse.lsp4mp.ls/src/test/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletionTest.java
```diff
@@ -365,6 +365,20 @@ public void completionBetweenIncompletePropertyNameAndEquals() throws BadLocatio
...
+				c("quarkus.http.cors.headers", "quarkus.http.cors.headers=existing", r(0, 0, 26)));
+	}
+
+	@Test
+	public void completionBetweenPropertyNameAndSpaceWithValue() throws BadLocationException {
+		String value = "quarkus.http.cors| =existing";
+		testCompletionFor(value, false,
+				c("quarkus.http.cors.headers", "quarkus.http.cors.headers=existing", r(0, 0, 27)));
+	}
+
```
rgrunber, line 381: Can you also add the following testcase :

```
@Test
public void completionBetweenPropertyNameAndSpaceWithBoolean() throws BadLocationException {
	String value = "quarkus.banner.en|=existing";
	testCompletionFor(value, false,
			c("quarkus.banner.enabled", "quarkus.banner.enabled=existing", r(0, 0, 26)));
	testCompletionFor(value, true,
			c("quarkus.banner.enabled", "quarkus.banner.enabled=${1|false,true|}", r(0, 0, 26)));
}

For now this just describes the current functionality. Maybe we can improve it in the future.
```

rgrunber (approved):

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
rgrunber: We can live with this for now.

PR #271, opened 2022-07-07
Fix NPE when hovering in properties file

### LINKED ISSUE #265: Hovering property value fails with NPE
Hovering a property value in a properties file fails due to an NPE:

```
[Error - 11:14:38 AM] Request textDocument/hover failed.
  Message: Internal error.
  Code: -32603 
java.util.concurrent.CompletionException: java.lang.NullPointerException: Cannot invoke "org.eclipse.lsp4j.jsonrpc.messages.Either.isLeft()" because "contents" is null
```

To reproduce this issue, create a MicroProfile project with a String config property "asdf" and a `microprofile-config.proeprties` file. In the properties file, add the following:

`asdf=hjkl`

Hover over `hjkl` and an error will be logged

datho7561:
lsp4j does not support serializing `Hover` without `contents`, so return null instead of a blank hover.

Closes #265


datho7561:
@rgrunber If you have time, could you please review this PR?

rgrunber (approved):
Fixes the issue for me.

PR #273, opened 2022-07-12
Do not give CodeActions for stale diagnostics

### LINKED ISSUE #272: Language Server attempts to calculate CodeAction for stale diagnostics
In https://github.com/eclipse/lsp4mp/commit/1d0377596b08e453c609bf5d0a0451af553a1e32, we delay revalidating a file, which means that we delay pushing new diagnostics until the user stops modifying the document, leaving stale diagnostics in the document.

However, according to the LSP, https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionContext, the client is free to request code actions to address these diagnostics, even after edits have been made to the document that might make them no longer relevant.

This is an issue, since our code to generate CodeActions currently assumes that the diagnostics it receives from the client are applicable to the current state of the document. When this assumption doesn't hold, bad CodeActions may be generated, or CodeAction generation may fail with an Exception. For instance, given this property file:

```properties
invalid.property|=value
```

If you press backspace with your cursor at the `|`, when you delete the last remaining `i` of `invalid.property`, an Exception will occur, since the language server will attempt to calculate the CodeAction to ignore the fact that the property `i` doesn't exist, however the property `i` is no longer present in the document.

We need some strategy to avoid calculating CodeActions when the diagnostics that are sent from the client don't match up with the current state of the document.

datho7561:
Fixes #272

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/commons/ValidatorDelayer.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/commons/ValidatorDelayer.java
@@ -83,4 +83,16 @@ public class ValidatorDelayer<T extends TextDocument> {
 			}
 		}
 	}
+
+	/**
+	 * Returns true if the document has a revalidation pending and false otherwise.
+	 *
+	 * @param uri the uri of the document to check
+	 * @return true if the document has a revalidation pending and false otherwise
+	 */
+	public boolean isRevalidating(String uri) {
+		synchronized (pendingValidationRequests) {
+			return pendingValidationRequests.containsKey(uri);
+		}
+	}
 }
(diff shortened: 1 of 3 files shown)


datho7561:
I'm going to try to write a test

.vscode/launch.json
```diff
@@ -0,0 +1,11 @@
+{
```
datho7561, line 1: I can remove this from this PR

rgrunber:
Then we can just evaluate the change based on trying it out.

rgrunber (approved):
Change looks good to me. I no longer see the error.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/commons/ValidatorDelayer.java
```diff
@@ -83,4 +83,16 @@ public void cleanPendingValidation(String uri) {
...
+
+	/**
+	 * Returns true if the document has a revalidation pending and false otherwise.
+	 *
+	 * @param uri the uri of the document to check
+	 * @return true if the document has a revalidation pending and false otherwise
+	 */
+	public boolean isRevalidating(String uri) {
+		synchronized (pendingValidationRequests) {
+			return pendingValidationRequests.get(uri) != null;
```
angelozerr, line 95: I think it  should be cleaner to use Map.containsKey()

PR #274, opened 2022-07-14
Remove unnecessary Gson dependency in pom file.

rgrunber:
- com.google.gson will come from the o.e.jdt.ls.tp target so there is no
  need to specify it as a dependency in the pom


datho7561 (approved):
Although it's a little late, the language features still appear to be working after this change. Good catch, Roland!

PR #276, opened 2022-07-20
Fix for excessive property definition display on hover

### LINKED ISSUE #215: Definition hover displays properties defined below target property
As per https://github.com/redhat-developer/quarkus-ls/pull/458#issuecomment-981747909, hovering a property definition usinf `Ctrl + Hover` will display properties defined below the target property.

E.g. if you have an `application.properties` file as follows:
```
greeting.first=
greeting.message=
greeting.second=
```
The definition hover in the java file will display the following:
Screenshot from 2021-11-29 10-31-55

AlexXuChen:
Fix for excessive property definition display on hover.
image

Fixes #215 

Fix made as per https://github.com/eclipse/lsp4mp/issues/215#issuecomment-984714139

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
@@ -152,8 +152,7 @@ public class JavaFileTextDocumentService extends AbstractTextDocumentService {
 				return null;
 			}
 			boolean canSupportMarkdown = true;
-			boolean snippetsSupported = sharedSettings.getCompletionCapabilities()
-					.isCompletionSnippetsSupported();
+			boolean snippetsSupported = sharedSettings.getCompletionCapabilities().isCompletionSnippetsSupported();
 			CompletionList list1 = new CompletionList();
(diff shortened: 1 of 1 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
```diff
@@ -290,8 +289,9 @@ public CompletableFuture<Either<List<? extends Location>, List<? extends Locatio
 									return location;
 								}).collect(Collectors.toList());
 						if (isDefinitionLinkSupport()) {
-							// I don't understand
-							// return Either.forRight(locations);
+							// Uncommented in
+							// https://github.com/redhat-developer/quarkus-ls/issues/499#issuecomment-1021555227
```
rgrunber, line 293: I think you meant to reference https://github.com/eclipse/lsp4mp/issues/215#issuecomment-984714139 , right ?

AlexXuChen: Ah whoops, yes I do

AlexXuChen:
Seems hover works properly without this change

angelozerr:
This  change must be done for havong a proper hyperlink when a definition is processeur.

PR #279, opened 2022-07-29
Use `computeModelAsyncCompose` from quarkus-ls

### LINKED ISSUE #257: Copy over `ModelTextDocuments#computeModelAsyncCompose` from quarkus-ls into commons package
See https://github.com/redhat-developer/quarkus-ls/pull/659/files?w=1#diff-8ffdd77369ed4ef0561d74cd65bfffd95327217e65143805720c33a137b991dd

This helper method is very helpful when implementing the `TextDocumentService` for an lsp4j-based language server, so it makes sense to include in the commons package.

datho7561:
This method makes it easier to implement go to definition in properties files. Includes the recent bug fix to the method (see redhat-developer/quarkus-ls#679)

Closes #257

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/commons/ModelTextDocuments.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/commons/ModelTextDocuments.java
@@ -15,11 +15,13 @@ package org.eclipse.lsp4mp.ls.commons;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.function.BiFunction;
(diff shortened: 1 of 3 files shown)


angelozerr:
I think we should do the same thing in https://github.com/eclipse/lsp4mp/blob/4adecbd009fdff3d53293fcf92bea7c318108244/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaTextDocuments.java#L130 no?

datho7561:
Good catch!

datho7561:
@angelozerr if you have some time, could you please take a look at this PR?

angelozerr:
LGTM, thanks @datho7561 !

Issue #207, opened 2021-10-20
Support for set of properties file


angelozerr:
Today completion, validation, definition works inside a given properties file but it should take care of another outside properties file.

# Completion sample:

 * In `.env` file:

`BDUSER=sa`
  
 * In `microprofile-config.properties` when completion is processed we should see `DBUSER`

# Definition sample:

 * In `.env` file:

`BDUSER=sa`
  
 * In `microprofile-config.properties` when definition is processed for DBUSER
 
 `DBUSER = sb`

it should open .env and select DBUSER.

# Validation sample:

 * In `micoroprofile-config.properties` file:

`quarkus.application.name=A`
  
 * In `application.properties` file:

`quarkus.application.name=B`

we should have a duplicate properties OR better having a diagnostic in quarkus.application.name inside microprofile-config.properties with `DiagnosticTag.Deprecated` to explain that the property is overriden by application.properties.

angelozerr:
When Set of properties files will be implemented, it will open the door to fix https://github.com/redhat-developer/vscode-quarkus/issues/192

PR #283, opened 2022-08-19
Use SmallRye Expression library to resolve property expressions

### LINKED ISSUE #241: Property evaluation should support the env variable notation
Originally posted in https://github.com/redhat-developer/intellij-quarkus/issues/472

When evaluating property value the notation ${ENVNAME:default} should be supported.

This breaks the JAX-RS codelens

datho7561:
Create a target platform so that smallrye-common-expression and future dependencies can easily be added to the jdt extension project.

Closes #241


datho7561:
See redhat-developer/vscode-microprofile#107 for relevant changes to a VS Code client that consumes lsp4mp.

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="Running Platform">
+	<locations>
+		<location type="Target" uri="https://raw.githubusercontent.com/eclipse/eclipse.jdt.ls/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target" />
```
datho7561, line 5: Is there a better way to reference the JDT-LS target? Would it be better to reference a tagged release?

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -0,0 +1,52 @@
...
+			<dependencies>
+				<dependency>
+					<groupId>io.smallrye.common</groupId>
+					<artifactId>smallrye-common-constraint</artifactId>
+					<version>1.12.0</version>
+				</dependency>
+			</dependencies>
+		</location>
+		<location includeSource="true" missingManifest="generate" type="Maven">
+			<feature id="org.jboss.logging" label="JBoss Logging" provider-name="JBoss" version="1.0.0.qualifier" />
```
datho7561, line 37: It looks like I could also reference a slightly out of date copy of the library from orbit. Would that be better?

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="Running Platform">
+	<locations>
+		<location type="Target" uri="https://raw.githubusercontent.com/eclipse/eclipse.jdt.ls/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target" />
```
datho7561, line 5: Is there a better way to reference this target file? Would it be better to reference the target file from a tagged release?

rgrunber: The only disadvantage to this is rebuilding a tag would never work as expected as it would always use the latest :\ .

So either we always refer to a tagged release, or in the case where there are breaking changes, some commit reference.

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
rgrunber, line 1: Would it be possible to add a p2 repository entry for https://download.eclipse.org/jdtls/milestones/1.14.0/repository/ ? Specifically, just `org.eclipse.jdt.ls.core` is needed. I think this would eliminate having the pom file entry for that repository.

Also by the looks of it.. PDE doesn't support adding the Maven coordinates to the target platform :disappointed:  I was really hoping just adding something like :

**javaConfig.json**
```
{
    "projects": [
        "microprofile.jdt/org.eclipse.lsp4mp.jdt.core",
        "microprofile.jdt/org.eclipse.lsp4mp.jdt.test"
    ],
   "targetPlatform": "microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target"
}
```

could also build the whole thing in VSCode !

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -0,0 +1,51 @@
...
+<?pde version="3.8"?>
+<target name="Running Platform">
+	<locations>
+		<location type="Target" uri="https://raw.githubusercontent.com/eclipse/eclipse.jdt.ls/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target" />
+		<location includeSource="true" missingManifest="generate" type="Maven">
+			<feature id="io.smallrye.common.expression" label="Smallrye Commons Expression" provider-name="Smallrye" version="1.0.0.qualifier" />
```
rgrunber, line 7: Is the `feature` element here really necessary ? All it seems to be doing is grouping the bundles, but it doesn't force the Bundle-SymbolicName or anything like that as far as I can see. Is it not enough to list all 4 of the below Maven artifacts under one `<dependencies>` element ?

Also, if you import and activate the target file in Eclipse IDE, you'll see it adds some elements to the XML (eg. `<type>jar</type>` inside <dependency> elements, and `includeDependencyDepth="none"` as an attribute of `<location>` (note that `infinite` is an option which is nice to know).

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/PropertiesConfigSource.java
```diff
@@ -46,7 +47,10 @@ public PropertiesConfigSource(String configFileName, int ordinal, IJavaProject j
 	@Override
 	public String getProperty(String key) {
 		Properties properties = getConfig();
-		return properties != null ? properties.getProperty(key) : null;
+		if (properties == null) {
+			return null;
+		}
+		return new PropertyValueExpander(properties).getValue(key);
```
angelozerr, line 53: I think expansion should be done: 

 * only if value is not an expression (the value doesn't contains `${`).
 * done just one time. Perhaps you could store the expanded value in the properties instance once it is expaned (but what about when expansion throw an error?)
 * it requires some test, but expand should perhaps use the scope of properties file (application.properties and meta-config.properties file)?

angelozerr:
@jeffmaury as the original issue comes from IJ, I think it should be nice to review the PR too if you have time.

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
datho7561: I looked into this, it looks like by using `includeDependencyDepth="infinite"`, I can get rid of the references to the dependencies. I removed the `feature` element as well.

datho7561: Eclipse didn't do this automatically for me though?

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
datho7561: Okay, I took out the pom file entry and added it to the target platform

rgrunber:
As a note, for the 3rd party libraries :

- `org.jgrapht.jgrapht-core 1.4.0` via. Orbit : https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23664
- `smallrye-common-expression 1.12.0` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/io.smallrye.common/smallrye-common-expression/1.12.0 (as per https://www.eclipse.org/projects/handbook/#ip-clearlydefined-using )
-  `smallrye-common-constraint 1.12.0` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/io.smallrye.common/smallrye-common-constraint/1.12.0
- `smallrye-common-function 1.12.0` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/io.smallrye.common/smallrye-common-function/1.12.0
- `org.jboss.logging 3.4.3.Final` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/org.jboss.logging/jboss-logging/3.4.3.Final

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/PropertiesConfigSource.java
datho7561: Okay. It should now work with multiple properties files. (i.e. it can resolve properties that are only given a value in a different property file). I had to modify the API for IConfigSource (there wasn't a way around doing it that was obvious to me).

datho7561: It seems like this API change doesn't affect quarkus-ls

datho7561:
> As a note, for the 3rd party libraries :
> 
> * `org.jgrapht.jgrapht-core 1.4.0` via. Orbit : https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23664
> * `smallrye-common-expression 1.12.0` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/io.smallrye.common/smallrye-common-expression/1.12.0 (as per https://www.eclipse.org/projects/handbook/#ip-clearlydefined-using )
> * `smallrye-common-constraint 1.12.0` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/io.smallrye.common/smallrye-common-constraint/1.12.0
> * `smallrye-common-function 1.12.0` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/io.smallrye.common/smallrye-common-function/1.12.0
> * `org.jboss.logging 3.4.3.Final` via. ClearlyDefined : https://clearlydefined.io/definitions/maven/mavencentral/org.jboss.logging/jboss-logging/3.4.3.Final

Also:
* `org.jheaps.jheaps 0.14.0.v20220411-2343` via Orbit: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23664

datho7561:
It'd be nice to consume `smallrye-common-expression` in `microprofile.ls` so that we can properly resolve complex syntax like `hjkl=${${asdf}}` and `$hjkl={asdf:DEFAULT_VALUE}` properly in the support for microprofile config property files (i.e. hover, completion in properties files). I will try to work on this tomorrow; I think that it's important for this PR, since it makes the behaviour consistent between Java files and properties files.

angelozerr:
Today on properties side expression evaluation is done on hover. But I would like to show result of expression évaluation with inlay hints. In ostherwise in this usecase évaluation will be done on each key stroke.

I agree with you that for consistent it is good but my fear is that when we will evaluate expression for inlay hint we will create some mp commons context on each key stroke which could be not good for performance.

angelozerr:
> I think I was testing the wrong thing, via. properties file. When I try to test in Java just for the basic hover support in @ConfigProperty, it's not working for me.

You need to save the properties file before hovering something. The hover, codelens don't use the properties file coming from src but coming from target folder (for maven). It is the reason why you need to save file in order build is done and copy/paste the properties file in target folder.

datho7561:
> ... we will create some mp commons context on each key stroke ...

If I don't understand correctly, please correct me. We should be able to resolve the properties within the properties file without needing to contact the JDT LS component, as long as we don't try to get values from other config sources. My idea is just to handle more advanced syntax like nested property expressions (`${${asdf}}`), which we don't do yet.

> I don't know if it that, ...

Good point, although when Roland was testing, this feature did not work as expected (I pushed a fix at the end of the day yesterday)

angelozerr:
> If I don't understand correctly, please correct me. We should be able to resolve the properties within the properties file without needing to contact the JDT LS component

Indeed I would like to evaluate expression (like we do today) on MP LS side only. We need to improve our properties parser to support advanced syntax like `${${asdf}}`

>  as long as we don't try to get values from other config sources. 

Indeed the MP LS must be improved with https://github.com/eclipse/lsp4mp/issues/207 Supporting a set of properties will open the door to support advanced features like supporting definition in expression to jump to an another properties file, like managing validation for dupplicate property, etc and to evauluate too expression where content can become from an another file.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/utils/PropertyValueExpander.java
```diff
@@ -0,0 +1,241 @@
...
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.commons.utils;
```
angelozerr, line 14: Dont use commons package which means that MP LS have the same classes. You could use `org.eclipse.lsp4mp.jdt.internal.core`

datho7561: mp-ls does have the same class

datho7561:
I pushed an amend so that the microprofile.ls component and the microprofile.jdt component use the same code in order to resolve property expressions. I removed the dependency on JGraphT, since Roland pointed out that Guava is actually available in both microprofile.jdt and microprofile.ls. I refactored the properties file model so that there are no "resolve" methods in these classes. The idea is you can use a utility method to resolve the values given the property model.

As well, the CI fails with "Unknown host" when trying to down load the JDT-LS target platform file. If this becomes a reoccuring problem, it's probably a good idea to look into a different way to define the target platform.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/utils/PropertyValueExpander.java
angelozerr: Ok thanks for the info.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/utils/PropertyValueExpander.java
```diff
@@ -0,0 +1,241 @@
...
+public class PropertyValueExpander {
+
+	private static final Logger LOGGER = Logger.getLogger(PropertyValueExpander.class.getName());
+	private static final long REFERENCE_UPPER_BOUND = 1_000_000;
+
+	private final PropertyGraph propertyGraph;
+	private final Map<String, String> properties;
+	private final Map<String, ResolvedPropertyValueInformation> resolved;
+
+	public PropertyValueExpander(Map<String, String> properties) {
```
angelozerr, line 54: Having PropertyValueExpander in commons is very nice but it requires to build a temporary Map to consume it. For MP LS, it means that this temporary Map will be build on each key stroke which is not good for performance.

Instead of using Map<String, String> properties, I will try to introduce a `IConfigSourcePropertiesProvider` which will give the capability to get information that you need (keys, values). IConfigSourcePropertiesProvider implementation is a wrapper which be implemented internally with :

 * list of IConfigSource (on JDT side)
 * a PropertiesModel on MP LS side (that we will change into a list of PropertiesModel in the future).

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/utils/PropertyValueExpander.java
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Expands property expressions to a final resolved value, while avoiding a few
+ * pitfalls.
+ * 
+ * Checks for cycles before attempting expansion, and mitigates OOM due to
+ * Billion Laughs by counting the number of variable references.
+ * 
+ * @author datho7561
+ */
+public class PropertyValueExpander {
```
angelozerr, line 45: It seems that this PropertyValueExpander from JDT is not the same than PropertyValueExpander from MP LS.

datho7561: Okay, they should be identical now

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/utils/PropertyValueExpander.java
datho7561: Let me know what you think of my implementation.

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="Running Platform">
```
rgrunber, line 3: Maybe we can call this `LSP4MP Target Platform`

rgrunber:
Overall, seems to be working well for me, so I think we should be ready to merge if no other issues.

datho7561:
Rebased; I changed the behaviour of the codelens slightly. Instead of hiding the codelens when part of a property expression can't be resolved, it shows the unresolved value (eg. `${bad.value}`). @angelozerr if you have some time, let me know if you think this is okay.

Also it looks like loading the remote target platform in the build failed again.

angelozerr:
LGTM

angelozerr:
Very great job @datho7561 !

PR #288, opened 2022-09-15
Update Target Platform to 1.16.0-SNAPSHOT version of JDT-LS target.

rgrunber:
- Remove unnecessary reference to JDT-LS version, since repository references automatically get access to all p2 bundles


datho7561 (approved):
Works for me. Didn't find any references to `jdt.ls.version`. Thanks for fixing this, Roland!

PR #293, opened 2022-09-30
Show value when hovering unrecognized property

### LINKED ISSUE #285: Show value of property when hovering key even when it isn't recognized
Say we have a `microprofile-config.properties` file:

```properties
asdf=foo ${bar}
bar=${hjkl} World!
hjkl = Hello,
```

If the property `asdf` were defined somewhere in the application with `@ConfigProperty`, you can hover over the `asdf` before the `=`, and it will show the final resolved value. However, this is not supported for properties that aren't defined in the application. I think it could be helpful to show the final resolved value in these cases.

datho7561:
Closes #285


angelozerr:
Code is perfect and the feature works great. 

Now we can see the resolved expression value for property which doesn't exists:

image

Thanks @datho7561 !

PR #294, opened 2022-09-30
Add diagnostics for mp-reactive-messaging @Incoming/Outgoing annotation

### LINKED ISSUE #58: Diagnostics for mp-reactive-messaging @Incoming/Outgoing annotation
Validate that @Incoming/Outgoing annotation are not blank. See https://github.com/eclipse/microprofile-reactive-messaging/blob/62c9ed5dffe01125941bb185f1433d6307b83c86/api/src/main/java/org/eclipse/microprofile/reactive/messaging/Incoming.java#L95

JessicaJHee:
Peek 2022-09-30 15-10

Fixes #58 


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/reactivemessaging/java/MicroProfileReactiveMessagingASTValidator.java
```diff
@@ -0,0 +1,85 @@
...
+public class MicroProfileReactiveMessagingASTValidator extends JavaASTValidator {
+
+    private static final String BLANK_CHANNEL_NAME_MESSAGE = "The name of the consumed channel must not be blank.";
+
+    private static final Logger LOGGER = Logger.getLogger(MicroProfileReactiveMessagingASTValidator.class.getName());
+
+    @Override
+    public void initialize(JavaDiagnosticsContext context, List<Diagnostic> diagnostics) {
+        super.initialize(context, diagnostics);
+    }
```
rgrunber, line 59: If you're not doing anything special with the class, you don't have to declare and have it just call the base class. You can simply leave it out. Because you are inheriting from `JavaASTValidator`, anything that calls `initialize(..)` on your class here, will just get the base anyways.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/reactivemessaging/java/MicroProfileReactiveMessagingASTValidator.java
```diff
@@ -0,0 +1,85 @@
...
+        boolean adapted = JDTTypeUtils.findType(javaProject, INCOMING_ANNOTATION) != null
+                || JDTTypeUtils.findType(javaProject, OUTGOING_ANNOTATION) != null;
+        return adapted;
+    }
+
+    @Override
+    public boolean visit(SingleMemberAnnotation node) {
+        validateIncomingOutgoingAnnotation(node);
+        super.visit(node);
+        return true;
```
rgrunber, line 74: I don't think you need these. In fact your could return `false` here. Although I notice a few example in the code that do exactly the same thing.

`super.visite(node)` just calls the base class, which if you look at, does nothing other than return a value, which we're ignoring here.

The boolean return value from a visit call indicates whether the caller should continue to visit the children of this element. This might be useful if some of the elements you care about are children of `SingleMemberAnnotation`. In our case, we know we won't find anything there, so we may as well return `false` and not traverse any deeper. Usually the reason to do this is performance.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/reactivemessaging/java/MicroProfileReactiveMessagingASTValidator.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private void validateIncomingOutgoingAnnotation(SingleMemberAnnotation node) {
+        if (node.getValue().toString().length() <= 2) {
```
rgrunber, line 78: Almost, but have a look at what the returned `Expression` from getValue() is. We know it needs to be a `String`. It might be any other type, but Java should handle the validation for us in those cases.

So all we need to do here is confirm the value is some type that represents a String. You might have to debug a correct example to figure out what that is.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/reactivemessaging/java/MicroProfileReactiveMessagingASTValidator.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public boolean visit(NormalAnnotation node) {
+        validateIncomingOutgoingAnnotation(node);
+        return false;
+    }
+
+    private void validateIncomingOutgoingAnnotation(Annotation node) {
+        try {
+            Expression expression = AnnotationUtils.getAnnotationMemberValueExpression(node, ATTRIBUTE_VALUE);
+            if (expression != null && ((StringLiteral) expression).getLiteralValue().isBlank()) {
```
rgrunber, line 87: Nice. I saw `isBlank()` initially but thought the annotations were not meant to emit diagnostics on things like "     ". However it makese sense that they would. They do say it must not be blank so this is a nice improvement :)

rgrunber, line 87: An `Expression` can be things other than a `StringLiteral`. While it's true that the JDT will handle those cases for us, our code still gets called, so we should handle this gracefully. Currently if you try `@Incoming(33)` or something like that, you should see that it gets caught by https://github.com/eclipse/lsp4mp/blob/b8c6d0651ae7c59f137a59a240fde3f40874a2ca/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/validators/MultiASTVisitor.java#L72

rgrunber (approved):

PR #302, opened 2022-10-19
Provide support for config_ordinal property

### LINKED ISSUE #289: Support for the config_ordinal property in microprofile-config.properties
The `config_ordinal` is flagged as an "unknown property":
image
but defined in the spec: https://download.eclipse.org/microprofile/microprofile-config-1.4/microprofile-config-spec.html#_manually_defining_the_ordinal_of_a_built_in_configsource

JessicaJHee:
Screenshot from 2022-10-19 16-17-18

Fixes #289 

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
@@ -100,6 +100,12 @@
       <validator class="org.eclipse.lsp4mp.jdt.internal.config.java.MicroProfileConfigASTValidator" />
    </extension>
 
+   <!-- Microprofile Config Static Property support -->
+
+   <extension point="org.eclipse.lsp4mp.jdt.core.propertiesProviders">
+      <provider class="org.eclipse.lsp4mp.jdt.internal.config.properties.MicroProfileConfigStaticPropertyProvider" />
+   </extension>
+
    <!-- Microprofile Context Propagation support -->
 
    <extension point="org.eclipse.lsp4mp.jdt.core.propertiesProviders">
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/properties/MicroProfileConfigStaticPropertyProvider.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/properties/MicroProfileConfigStaticPropertyProvider.java
@@ -1,0 +1,39 @@
+/*******************************************************************************
+* Copyright (c) 2022 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.jdt.internal.config.properties;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.lsp4mp.jdt.core.AbstractStaticPropertiesProvider;
+import org.eclipse.lsp4mp.jdt.core.MicroProfileCorePlugin;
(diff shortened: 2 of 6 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/configordinal/properties/ConfigOrdinalPropertyProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class ConfigOrdinalPropertyProvider extends AbstractStaticPropertiesProvider {
+
+	public ConfigOrdinalPropertyProvider() {
+		super(MicroProfileCorePlugin.PLUGIN_ID, "/static-properties/mp-configordinal-metadata.json");
+	}
+
+	@Override
+	protected boolean isAdaptedFor(SearchContext context, IProgressMonitor monitor) {
+		return true;
```
rgrunber, line 34: Would it make sense to check for [`org.eclipse.microprofile.config.Config`](https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/Config.java) as a sanity check ?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/configordinal/properties/ConfigOrdinalPropertyProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import org.eclipse.lsp4mp.jdt.core.AbstractStaticPropertiesProvider;
+import org.eclipse.lsp4mp.jdt.core.MicroProfileCorePlugin;
+import org.eclipse.lsp4mp.jdt.core.SearchContext;
+
+/**
+ * Config ordinal property provider.
+ * 
+ * @see https://download.eclipse.org/microprofile/microprofile-config-1.4/microprofile-config-spec.html#_manually_defining_the_ordinal_of_a_built_in_configsource
+ */
+public class ConfigOrdinalPropertyProvider extends AbstractStaticPropertiesProvider {
```
angelozerr, line 26: I will rename this class to `MicroProfileConfigPropertyProvider`  to embed future properties for the microprofile-config module.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/configordinal/properties/ConfigOrdinalPropertyProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+	public ConfigOrdinalPropertyProvider() {
+		super(MicroProfileCorePlugin.PLUGIN_ID, "/static-properties/mp-configordinal-metadata.json");
```
angelozerr, line 29: I will rename `mp-configordinal-metadata.json` to `mp-config-metadata.json`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/configordinal/properties/ConfigOrdinalPropertyProvider.java
angelozerr: Yes I think it can be a good idea. Ideally we should check the MP version, but we have no this feature. One idea that I have is to manage a `since` JSON field in the *.medata.json and use it to check if it matches with the Java library version.

But we could do that in an another PR.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/configordinal/properties/ConfigOrdinalPropertyProvider.java
JessicaJHee: We have a class with the same name here: https://github.com/eclipse/lsp4mp/blob/51d5e8715c27a05c439fd84fe10e9cf3fd7b8bef/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/properties/MicroProfileConfigPropertyProvider.java#L52 

But it doesn't extend `AbstractStaticPropertiesProvider` which works with static properties. Would `MicroProfileConfigStaticPropertyProvider` make sense? And I could move this class in `eclipse/lsp4mp/jdt/internal/config/properties`

PR #308, opened 2022-11-08
Use FQN when checking annotation type in `AnnotationUtils`

### LINKED ISSUE #304: MicroProfileConfigASTValidator handles @ConfigProperties in a wrong way
`MicroProfileConfigASTValidator` tries to locate `@ConfigProperties` but it uses only the simple class name and not the fully qualified class name.
As a result, it also works for Quarkus `@ConfigProperties` but it should located into the Quarkus specific part and it may wrongly process `@ConfigProperties` annotations from other packages.

datho7561:
Fixes #304

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java
@@ -29,6 +29,8 @@ import org.eclipse.jdt.core.ITypeRoot;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.dom.Annotation;
 import org.eclipse.jdt.core.dom.Expression;
+import org.eclipse.jdt.core.dom.IAnnotationBinding;
+import org.eclipse.jdt.core.dom.ITypeBinding;
 import org.eclipse.jdt.core.dom.MemberValuePair;
 import org.eclipse.jdt.core.dom.NormalAnnotation;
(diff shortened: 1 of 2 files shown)


datho7561:
This is the quick version of the fix that addresses just the issue that Jeff mentioned. However, after making this change, I found some bugs with collecting and validating properties coming from `@ConfigProperties`.

rgrunber:
Yeah, my suspicion was you may uncover some other places we do things in a similarly wrong way :\

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java
```diff
@@ -183,7 +184,8 @@ private static boolean isMatchAnnotationFullyQualifiedName(IAnnotation annotatio
 	 *         false otherwise.
 	 */
 	public static boolean isMatchAnnotation(Annotation annotation, String annotationName) {
-		return annotationName.endsWith(annotation.getTypeName().getFullyQualifiedName());
+		IAnnotationBinding binding = annotation.resolveAnnotationBinding();
+		return annotationName.endsWith(binding.getAnnotationType().getQualifiedName());
```
rgrunber, line 188: You can even do : `annotation.resolveTypeBinding().getQualifiedName()` though I do like your approach because it goes more to what the type happens to be.

Should we guard off `binding.getAnnotationType()` ? If I look into the implementation, it can return `null`.

datho7561: Sure, I'll add a check for that

rgrunber (approved):

PR #309, opened 2022-11-08
Add REST client CodeLens for additional annotations

### LINKED ISSUE #94: REST client CodeLens only shows up for GET annotations.
Currently the MicroProfile REST Client CodeLens only show up for paths annotated with `@GET`.

I think it makes sense to display them for other annotations like POST or PUT even though they cannot be accessed from the browser. (If possible maybe we could make CodeLens for POST/PUT etc. requests not clickable?) 

IE. if u have a rest client with just a POST endpoint, you might still want the CodeLens to appear so you can quickly see the full URL that is configured by your mp-config.properties file.

JessicaJHee:
Fixes #94

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
@@ -15,6 +15,7 @@ package org.eclipse.lsp4mp.jdt.internal.jaxrs.java;
 
 import static org.eclipse.lsp4mp.jdt.core.jaxrs.JaxRsUtils.createURLCodeLens;
 import static org.eclipse.lsp4mp.jdt.core.jaxrs.JaxRsUtils.getJaxRsPathValue;
+import static org.eclipse.lsp4mp.jdt.core.jaxrs.JaxRsUtils.isClickableJaxRsRequestMethod;
 import static org.eclipse.lsp4mp.jdt.core.jaxrs.JaxRsUtils.isJaxRsRequestMethod;
 import static org.eclipse.lsp4mp.jdt.core.utils.JDTTypeUtils.overlaps;
 import static org.eclipse.lsp4mp.jdt.internal.jaxrs.JaxRsConstants.JAVAX_WS_RS_PATH_ANNOTATION;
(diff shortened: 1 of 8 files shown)


rgrunber:
test this please.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
```diff
@@ -133,7 +133,7 @@ private static void collectURLCodeLenses(IJavaElement[] elements, String rootPat
 				// @DELETE, @PUT
 				// JAX-RS
 				// annotation
-				if (isJaxRsRequestMethod(method) && Flags.isPublic(method.getFlags())) {
+				if (isClickableJaxRsRequestMethod(method) && Flags.isPublic(method.getFlags())) {
```
datho7561, line 136: What do you think about also providing the unclickable urls for the application's own REST endpoints in this PR? When I read the issue initially, that's what I thought it was talking about.

datho7561 (approved):
This works well, and the code looks good! I think we are going to hold off on merging this until after the release, although we should discuss that.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/jaxrs/java/JaxRsCodeLensParticipant.java
JessicaJHee: I think it makes sense to do that as well. @rgrunber what do you think?

rgrunber:
Change works well for me. Do we care about `HEAD` in the event a user decides to customize ? or is that basically overkill.

jaxrs-head-support

JessicaJHee:
> Change works well for me. Do we care about `HEAD` in the event a user decides to customize ? or is that basically overkill.

I didn't include it initially because I thought it wasn't very useful to show, but I can also add it to be consistent since it is allowed?

rgrunber:
> > Change works well for me. Do we care about `HEAD` in the event a user decides to customize ? or is that basically overkill.
> 
> I didn't include it initially because I thought it wasn't very useful to show, but I can also add it to be consistent since it is allowed?

If it's easy enough, I would add it. It also isn't clickable, mainly because I don't think a web browser would render it.

rgrunber (approved):

PR #311, opened 2022-11-09
Fix inlay hint + definition when project returns empty properties.

angelozerr:
Fix inlay hint + definition when project returns empty properties.


angelozerr:
I have written test for inlay hint. 

I need to write test for definition now.

angelozerr:
It should be good now.

datho7561 (approved):
Looks good! I can't think of a way to test this manually, since even in non-microprofile projects the `config_ordinal` property is present (this is a bug, I think it should depend on microprofile-config being present).

datho7561:
See #312

PR #318, opened 2022-11-18
Use `JavadocContentAccess2`

### LINKED ISSUE #297: Use `JavadocContentAccess2` for better Markdown hover documentation
eclipse.jdt.ls makes use of `JavadocContentAccess2` to provide its Markdown-formatted hover documentation ([here is its use in the hover implementation](https://github.com/eclipse/eclipse.jdt.ls/blob/c15fe895fdc3eaca1386ff1cebde36f3af63f733/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/HoverInfoProvider.java#L311)). It would be nice to use the same method in JDTUtilsLSImpl.

datho7561:
Closes #297

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/properties/MicroProfileConfigPropertiesProvider.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/properties/MicroProfileConfigPropertiesProvider.java
@@ -33,8 +33,10 @@ import org.eclipse.jdt.core.IField;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IType;
 import org.eclipse.jdt.core.JavaModelException;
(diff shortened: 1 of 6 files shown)


rgrunber:
Would just be good to have an idea of what regresses as a result of this and what we gain.

datho7561:
I think I might need to investigate further. When I use `JavadocContentAccess2`, the `{@link` tags try to point to `jdt://` for references to binary classes, which doesn't end up as a link in the vscode UI for some reason. I don't know why the test here doesn't reflect that, though.

datho7561:
See https://github.com/redhat-developer/vscode-java/issues/2810. I'll add a test here that demonstrates that `JavadocContentAccess2` generally gives better hover documentation results.

rgrunber:
Good catch on the javadoc extraction.

rgrunber:
Looks like this is causing https://ci.eclipse.org/lsp4mp/job/lsp4mp/job/master/186/testReport/org.eclipse.lsp4mp.jdt.core.faulttolerance.properties/MicroProfileFaultTolerancePropertiesTest/Build_LSP4MP_JDT_LS_extension___microprofileFaultTolerancePropertiesTest/ but worked for me locally.

PR #328, opened 2022-12-02
Upversion to 0.7.0-SNAPSHOT

datho7561:

### DIFF
--- a/microprofile.jdt/pom.xml
+++ b/microprofile.jdt/pom.xml
@@ -3,14 +3,13 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.eclipse.lsp4mp</groupId>
   <artifactId>parent</artifactId>
-  <version>0.6.0-SNAPSHOT</version>
+  <version>0.7.0-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>microprofile.jdt.ls :: parent</name>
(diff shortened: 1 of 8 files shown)


rgrunber:
I would just make the commit message/title reference `0.7.0-SNAPSHOT`.

microprofile.jdt/pom.xml
```diff
@@ -3,14 +3,14 @@
...
-  <version>0.6.0-SNAPSHOT</version>
+  <version>0.7.0-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>microprofile.jdt.ls :: parent</name>
   <description>microprofile.jdt.ls parent</description>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <project.version>0.6.0-SNAPSHOT</project.version>
+    <project.version>0.7.0-SNAPSHOT</project.version>
```
rgrunber, line 13: Just wondering if this is needed. Should the references not use 0.7.0-SNAPSHOT anyways because that's set at the top level ?

datho7561: Doesn't look like it, I'll remove

PR #329, opened 2022-12-06
Documentation on hover for properties in properties files

### LINKED ISSUE #321: Show config property documentation when hovering over the key in a properties file
eg. let's say you have `MyClass.java`, which receives an injected configuration property:

```java
public class MyClass {

  /**
   * This config configures this class.
   */
  @ConfigProperty(name="my.config")
  int value;
}
```

Also, you have a properties file:

```
my.value=1000
```

If you hover over `my.value` in the properties file, it would be nice to show the Javadoc on the `int value` member of `MyClass`.

datho7561:
Adds a new command `microprofile/propertyDocumentation` to the JDT component to retrieve the documentation for a property. Invokes this command asynchonously when hovering over the property key in a properties file.

Closes #321


rgrunber:
Seems to be working well for me. Just to confirm. The MP LS receives the hover request, and sends a `microprofile/propertyDocumentation` request back to the client, which has registered a command (via `java.execute.workspaceCommand`) that delegates to the MP extension contained within JDT-LS ?

rgrunber (approved):
Change looks good so feel free to merge when ready.

datho7561:
Thanks for reviewing this, Roland!

PR #330, opened 2022-12-07
Use jakarta imports in snippets when possible

### LINKED ISSUE #229: Add support for jakarta namespace
MP 5 supports jakarta namespace aligning with Jakarta EE 9/9.1.  The LS should support the use of packages with the jakarta namespace.

Cc @kathrynkodama

### LINKED ISSUE #320: Snippets using Java EE 8 when project is using Jakarta EE9+
For example, the `mpreadiness` snippet will be filtered on the context `org.eclipse.microprofile.health.Readiness` but there isn't additional filtering to check the EE version. So if I am using EE9, the jaxrs namespace will still be inserted via the snippet even though my project is using EE9 with the jakarta.x namespace. 

image

datho7561:
If the jakarta versions of the javax classes referenced in the current snippets are on the classpath, hide the old snippet and display a new snippet that imports the jakarta classes instead.

Closes #229, closes #320


angelozerr:
IMHO I think it is really shame to duplicate snippets just to generate the proper import `jakarta.enterprise.context.ApplicationScoped` or `javax.enterprise.context.ApplicationScoped` according the classpath.

It should be really nice if we could have just one snippet and write:

`import ${2:eepackagename}.enterprise.context.ApplicationScoped;`

and replace `${2:eepackagename}` with javax or jakarta package name according to the classpath. We have already a mechanism which replace `${1:packagename}` with the proper package name. 

JDT LS side should fill the proper eepackagename.

datho7561:
> It should be really nice if we could have just one snippet and write:
> 
> `import ${2:eepackagename}.enterprise.context.ApplicationScoped;`

Thanks for the feedback. I took a look at it, and I also think what you suggested is a better approach. I'll look into redoing this.

datho7561:
Another issue is that we need to change not just the snippet content, but also the snippet prefix and description

PR #331, opened 2022-12-07
Update Target Platform to 1.19.0-SNAPSHOT version of JDT-LS target.

rgrunber:
- Update Tycho 2.7.2 to 3.0.1

Wouldn't mind having it confirmed in https://github.com/eclipse-tycho/tycho/pull/1217#issuecomment-1341534832 that the new error we're seeing is ok, and not harmful.

Another oddity I encountered : If I don't update Tycho from 2.7.2 -> 3.0.1, then there is a resolution error in tycho-compiler-plugin's validation :ghost: .

### DIFF
--- a/microprofile.jdt/pom.xml
+++ b/microprofile.jdt/pom.xml
@@ -10,7 +10,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <tycho.version>2.7.2</tycho.version>
+    <tycho.version>3.0.1</tycho.version>
     <tycho.extras.version>${tycho.version}</tycho.extras.version>
     <tycho.scmUrl>scm:git:https://github.com/eclipse/lsp4mp</tycho.scmUrl>
     <tycho.generateSourceReferences>true</tycho.generateSourceReferences>
@@ -24,7 +24,7 @@
     <jacoco.destFile>${project.build.directory}/jacoco.exec</jacoco.destFile>
     <sonar.jacoco.reportPath>${jacoco.destFile}</sonar.jacoco.reportPath>
     <surefire.timeout>600</surefire.timeout>
-    <jdt.ls.version>1.16.0-SNAPSHOT</jdt.ls.version>
+    <jdt.ls.version>1.19.0-SNAPSHOT</jdt.ls.version>
   </properties>
   <scm>
     <connection>scm:git:git@github.com:eclipse/lsp4mp.git</connection>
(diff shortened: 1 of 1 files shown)


microprofile.jdt/pom.xml
```diff
@@ -36,7 +36,7 @@
     <repository>
       <id>jdt.ls.p2</id>
       <layout>p2</layout>
-      <url>https://download.eclipse.org/jdtls/milestones/1.15.0/repository/</url>
+      <url>https://download.eclipse.org/jdtls/snapshots/repository/latest/</url>
```
datho7561, line 39: I think that we had an issue pointing to the snapshots repository in the past, since if we don't update the jdt.ls version then the build breaks when jdt.ls bumps to a new minor version. Do you remember something like this?

rgrunber: Maybe with https://repo.eclipse.org/content/repositories/jdtls-snapshots/org/eclipse/jdt/ls/org.eclipse.jdt.ls.tp/ ? Looks like everything prior to 1.16.0 is gone so if we referred to an old enough snapshot, and didn't update for a while, we might run into this ? I don't think we'd have an issue if JDT-LS released though. https://download.eclipse.org/jdtls/snapshots/repository/ I cleaned manually a while ago. and we still have a lot of the older snapshots there.

I know that for https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target#L5 , the issue we had is that snapshot artifacts are referred to by URL there, and they can sometimes be cleaned out so it was best to refer to a released one.

datho7561: Okay. I think I got this mixed up with https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target#L5

datho7561 (approved):
Looks good. Thanks Roland!

microprofile.jdt/pom.xml
rgrunber: Ok, it just hit me what else could happen. The JDT-LS target file refers to say https://download.eclipse.org/eclipse/updates/4.26-I-builds , which is valid until 4.26 is released (eg. today). After that there is no requirement to retain it (4.25-I-builds are gone). In the past 4.27-I-builds might not even exist yet, but currently it does, and we updated to it. If JDT-LS releases a version that refers to the I-builds of an older release, then once Eclipse releases that I-build, we basically have to jump to a newer JDT-LS.

The solution here is to just stay up to date though.

PR #335, opened 2023-01-04
When finding node at offset, prefer `PropertyKey` over `Assign`

datho7561:
when finding the node at the `|`:

```property
key|=value
```

return the property key node (`key`) instead of the assign node (`=`).

Note that this doesn't affect the behaviour of this case:

```property
key |= value
```

which will still return the assign node (`=`).

Fixes redhat-developer/quarkus-ls#323

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/Property.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/Property.java
@@ -170,21 +170,32 @@ public class Property extends Node {
 		if (key == null) {
 			return this;
 		}
-		if (key.getEnd() == -1) {
+		/*
+		 * eg.
+		 *
+		 * asdf|
+		 * asdf| = value
+		 * asdf|=value
+		 */
+		if (key.getEnd() == -1 || offset <= key.getEnd()) {
 			return key;
 		}
 		Node assign = getDelimiterAssign();
 		if (assign == null) {
 			return key;
 		}
-		if (assign.getStart() == offset) {
+		Node value = getValue();
+		/*
+		 * eg.
+		 *
+		 * asdf =|
+		 * asdf |= value
+		 * asdf =| value
+		 */
+		if (value == null || offset < value.getStart()) {
 			return assign;
 		}
-		if (offset >= assign.getStart()) {
-			Node value = getValue();
-			return value != null ? value.findNodeAt(offset) : assign;
-		}
-		return key;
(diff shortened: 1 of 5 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/Property.java
```diff
@@ -170,21 +170,18 @@ public Node findNodeAt(int offset) {
 		if (key == null) {
 			return this;
 		}
-		if (key.getEnd() == -1) {
+		if (key.getEnd() == -1 || offset <= key.getEnd()) {
```
angelozerr, line 173: Could you add please add a sample in a comment which correspond to this usecase.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/model/Property.java
```diff
@@ -170,21 +170,18 @@ public Node findNodeAt(int offset) {
...
 			return key;
 		}
 		Node assign = getDelimiterAssign();
 		if (assign == null) {
 			return key;
 		}
-		if (assign.getStart() == offset) {
+		Node value = getValue();
+		if (value == null || offset < value.getStart()) {
```
angelozerr, line 181: Could you add please add a sample in a comment which correspond to this usecase.

microprofile.jdt/org.eclipse.lsp4mp.jdt.tp/org.eclipse.lsp4mp.jdt.target
```diff
@@ -2,7 +2,7 @@
 <?pde version="3.8"?>
 <target name="LSP4MP Target Platform">
 	<locations>
-		<!--<location type="Target" uri="https://raw.githubusercontent.com/eclipse/eclipse.jdt.ls/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target"/>-->
+		<location type="Target" uri="https://raw.githubusercontent.com/eclipse/eclipse.jdt.ls/master/org.eclipse.jdt.ls.target/org.eclipse.jdt.ls.tp.target"/>
```
angelozerr, line 5: Remove the commented line

angelozerr:
Once https://github.com/eclipse/lsp4mp/pull/335/files#r1071966394 is fixed, I'm OK to merge it, it works like a charm.

angelozerr:
It works like a charm, thanks @datho7561 !

PR #338, opened 2023-01-05
Show/hide Java snippets based on cursor location

### LINKED ISSUE #108: Only show Java file snippets relevent to the context
Currently, class snippets, such as `mpliveliness` and `mpreadiness` are suggested anywhere in a Java file:

mpreadiness-inside-class

It might be helpful to:
* Only show snippets that create class definitions when the cursor isn't inside a class definition
* Only show snippets that add methods to a class when the cursor is inside a class definition

datho7561:
Snippet completion sends a request to the java language server to determine additional context based on the cursor location and content of the java file, then filters the snippets based on this context.

The context is fetched as a part of the existing `microprofile/java/completion` request.

This means that:
- snippets that generate class declarations (like `rest_class`, `mpreadiness`) only appear in empty files
- snippets that generate an annotation only appear when the cursor is in a class and before a declaration that the given annotation can be applied to.

This PR also removes code related to enabling/disabling the feature of getting the completion results directly from the JDT component.

Closes #108

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContextKind.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContextKind.java
@@ -1,0 +1,234 @@
+/*******************************************************************************
+* Copyright (c) 2023 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.commons;
+
+/**
+ * Represents the context where the cursor is in a Java file.
+ */
+public enum JavaCursorContextKind {
+
+	/**
+	 * The cursor is in a file that does not have a root type declaration.
+	 *
+	 * eg.
+	 * <code><br />
+	 * |<br />
+	 * </code>
+	 */
+	IN_EMPTY_FILE(1),
+
+	/**
+	 * The cursor is before a type declaration body, either at the root of a file or
+	 * within another class. The cursor is before any annotations on the type.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * |<br />
+	 * &commat;Inject<br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * }<br />
+	 * </code>
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * &commat;Inject<br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;private int memberVariable;<br />
+	 * &emsp;&emsp;&emsp;&emsp;|<br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Inject<br />
+	 * &emsp;&emsp;&emsp;&emsp;public static class MyChildClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;}<br />
+	 * }<br />
+	 * </code>
+	 */
+	BEFORE_CLASS(2),
+
+	/**
+	 * The cursor is in a type declaration body, and the next declaration in the
+	 * body is a method declaration. The cursor is before any annotations on the
+	 * method.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * public class MyClass {<br />
+	 * &emsp;&emsp;&emsp;&emsp;|<br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Deprecated<br />
+	 * &emsp;&emsp;&emsp;&emsp;public void myMethod() {<br />
+	 * &emsp;&emsp;&emsp;&emsp;}<br />
+	 * }<br />
+	 * </code>
+	 */
+	BEFORE_METHOD(3),
+
+	/**
+	 * The cursor is in a type declaration body, and the next declaration in the
+	 * body is a field declaration. The cursor is before any annotations on the
+	 * field.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * public class MyClass {<br />
+	 * &emsp;&emsp;&emsp;&emsp;|<br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Deprecated<br />
+	 * &emsp;&emsp;&emsp;&emsp;public String myString;<br />
+	 * }<br />
+	 * </code>
+	 */
+	BEFORE_FIELD(4),
+
+	/**
+	 * The cursor is before a type declaration body, either at the root of a file or
+	 * within another class. The cursor is somewhere within the annotation
+	 * declarations on the class.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * &commat;Inject|<br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * }<br />
+	 * </code>
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * &commat;Inject<br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;private int memberVariable;<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Inject|<br />
+	 * &emsp;&emsp;&emsp;&emsp;public static class MyChildClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;}<br />
+	 * }<br />
+	 * </code>
+	 */
+	IN_CLASS_ANNOTATIONS(5),
+
+	/**
+	 * The cursor is in a type declaration body, and the next declaration in the
+	 * body is a method declaration. The cursor is somewhere within the annotation
+	 * declarations on the method.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Deprecated|<br />
+	 * &emsp;&emsp;&emsp;&emsp;public void myMethod() {<br />
+	 * &emsp;&emsp;&emsp;&emsp;}<br />
+	 * }<br />
+	 * </code>
+	 */
+	IN_METHOD_ANNOTATIONS(6),
+
+	/**
+	 * The cursor is in a type declaration body, and the next declaration in the
+	 * body is a field declaration. The cursor is somewhere within the annotation
+	 * declarations on the field.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Deprecated|<br />
+	 * &emsp;&emsp;&emsp;&emsp;public String myString;<br />
+	 * }<br />
+	 * </code>
+	 */
+	IN_FIELD_ANNOTATIONS(7),
+
+	/**
+	 * The cursor is in a type declaration body, after all the declarations for the
+	 * type.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Deprecated<br />
+	 * &emsp;&emsp;&emsp;&emsp;public String myString;<br />
+	 * &emsp;&emsp;&emsp;&emsp;|<br />
+	 * }<br />
+	 * </code>
+	 */
+	IN_CLASS(8),
+
+	/**
+	 * None of the above context apply.
+	 *
+	 * eg.
+	 * <code>
+	 * <br />
+	 * package org.acme;<br />
+	 * <br />
+	 * public class MyClass {<br />
+	 * <br />
+	 * &emsp;&emsp;&emsp;&emsp;&commat;Deprecated<br />
+	 * &emsp;&emsp;&emsp;&emsp;public void myMethod() {<br />
+	 * &emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;|<br />
+	 * &emsp;&emsp;&emsp;&emsp;}<br />
+	 * }<br />
+	 * </code>
+	 */
+	NONE(2000);
+
+	private final int value;
+
+	private JavaCursorContextKind(int value) {
+		this.value = value;
+	}
+
+	public int getValue() {
+		return value;
+	}
+
+	public static JavaCursorContextKind forValue(int value) {
+		JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
+		if (value < 1 || value > allValues.length)
+			throw new IllegalArgumentException("Illegal enum value: " + value);
+		return allValues[value - 1];
+	}
+
+}
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.IBuffer;
 import org.eclipse.jdt.core.IClassFile;
 import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaElement;
@@ -27,6 +28,21 @@ import org.eclipse.jdt.core.IMethod;
 import org.eclipse.jdt.core.ISourceRange;
 import org.eclipse.jdt.core.ITypeRoot;
 import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.ASTVisitor;
+import org.eclipse.jdt.core.dom.AbstractTypeDeclaration;
+import org.eclipse.jdt.core.dom.AnnotationTypeDeclaration;
+import org.eclipse.jdt.core.dom.AnnotationTypeMemberDeclaration;
+import org.eclipse.jdt.core.dom.BodyDeclaration;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.EnumConstantDeclaration;
+import org.eclipse.jdt.core.dom.EnumDeclaration;
+import org.eclipse.jdt.core.dom.FieldDeclaration;
+import org.eclipse.jdt.core.dom.MethodDeclaration;
+import org.eclipse.jdt.core.dom.NodeFinder;
+import org.eclipse.jdt.core.dom.RecordDeclaration;
+import org.eclipse.jdt.core.dom.TypeDeclaration;
+import org.eclipse.jdt.internal.core.manipulation.dom.ASTResolving;
 import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.CodeLens;
 import org.eclipse.lsp4j.CompletionItem;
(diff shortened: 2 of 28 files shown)


datho7561:
Here's the relevant change for vscode-microprofile: https://github.com/redhat-developer/vscode-microprofile/pull/139

angelozerr:
Thanks so much to provide this kind PR. When I implemented Java snippet, my goal was to avoid trigerring JDT LS on each key stroke. With your snippet cursor context, you will consume JST LS command on each key stroke. What about the performance? @fbricon what do you think about that?

datho7561:
> ...my goal was to avoid trigerring JDT LS on each key stroke. With your snippet cursor context, you will consume JST LS command on each key stroke.

Yeah, my initial idea was just to check if the file is empty or not in order to avoid contacting JDT-LS. However, JDT-LS manages the contents of Java files. There might be unsaved changes to the file that only JDT-LS knows about, which affects what snippets should be returned.

Also, I don't think that the MicroProfile language server contacts JDT-LS on every keystroke. It should only contact it once whenever completion is opened. You can confirm this by opening completion, then typing something. If you open the language server trance, there should only be one completion request to the MicroProfile language server, and one response.

However, thinking through this again, I just realized an important limitation that I will need to address before this PR is ready: if a user has the following Java file:

```java
rest_class|
```

and they open completion at the `|`, the `rest_class` snippet won't be returned since the file isn't empty.

datho7561:
I addressed the limitation that I mentioned by checking if the root type exists instead of checking if the file is empty.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
```diff
@@ -370,6 +371,42 @@ public Hover hover(MicroProfileJavaHoverParams params, IJDTUtils utils, IProgres
...
+		ITypeRoot typeRoot = resolveTypeRoot(uri, utils, monitor);
+		if (typeRoot == null || typeRoot.findPrimaryType() == null) {
+			return JavaCursorContext.EMPTY_FILE;
+		}
+
+		Position completionPosition = params.getPosition();
+		int completionOffset = utils.toOffset(typeRoot.getBuffer(), completionPosition.getLine(),
+				completionPosition.getCharacter());
+
+		IJavaElement element = typeRoot.getElementAt(completionOffset);
```
angelozerr, line 395: I think you should use AST CompilationUnit instead of DOM CompilationUnit to be more precise. To get AST CompilationUnit, see https://github.com/eclipse/lsp4mp/blob/caf91c68e7f2fe27072f1f947e3c5476f3289d8f/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/AbtractJavaContext.java#L98 and after that you get the covered node https://github.com/eclipse/lsp4mp/blob/f5571c37f71799acd5c0a0a14b43b735bee83313/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/JavaCodeActionContext.java#L94

angelozerr:
If wa want to manage with more precision those snippet with smart behvaior, I think the JSON context should look like this:

```json
"context": {
   "type": "class|method|field|annotation",
   "when": {
     "typeOnClasspath": [
        "javax.ws.rs.GET",
        "jakarta.ws.rs.GET"
     ]
   }
}
```

 * `type` defines the Java generated content:
   * `class`  : generate the full content of class with import.
   * `method`  : generate a method. 
   *  `field`  : generate a field.
   *   `annotation`  : generate an annotation.  
 * `typeOnClasspath` display the snippet when the given Java type is in the classpath.

You can notice that there is no `cursorContext`, because we can determine the context according to the `type`:

 * when type is  `class`, we check that the file is empty (like you did)
 * when type is  `method`, `field`, we check that the cursor is insde a class but not inside a method body.
 * when type is  `annotation`, we can on the first step display those annotation before a class, method or field. BUT it will be not enough because some annotation are applied only for class,method, field and even parameter type. So my idea is to get the annotation target with JDT https://github.com/eclipse/microprofile-open-api/blob/8c8adaad12a7dd4cdd0bd12bb3967e60d62db98f/api/src/main/java/org/eclipse/microprofile/openapi/annotations/Operation.java#L34 when type arre returned on MP LS. But let's see this usecase in an another PR.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/plugin.xml
```diff
@@ -41,6 +41,7 @@
             <command id="microprofile/java/definition"/>
             <command id="microprofile/java/diagnostics"/>
             <command id="microprofile/java/hover"/>
+            <command id="microprofile/java/snippetCursorContext"/>
```
angelozerr, line 44: I will rename `snippetCursorContext` with `javaCursorContext`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
```diff
@@ -0,0 +1,31 @@
...
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.commons;
+
+/**
+ * Represents the context where the cursor is in a Java file.
+ */
+public enum JavaCursorContext {
```
angelozerr, line 19: I will rename the enum with `JavaCursorContextKind`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
```diff
@@ -0,0 +1,31 @@
...
+	EMPTY_FILE(1),
+	CLASS(2),
+	NONE_OF_THE_ABOVE(3);
+
+	int id;
+
+	private JavaCursorContext(int id) {
+		this.id = id;
+	}
+
```
angelozerr, line 30: Define a `forValue` static method that LSP4J will use to serialize this enum like https://github.com/eclipse/lsp4mp/blob/caf91c68e7f2fe27072f1f947e3c5476f3289d8f/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/ClasspathKind.java#L42

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
```diff
@@ -0,0 +1,31 @@
...
+package org.eclipse.lsp4mp.commons;
+
+/**
+ * Represents the context where the cursor is in a Java file.
+ */
+public enum JavaCursorContext {
+
+	EMPTY_FILE(1),
+	CLASS(2),
+	NONE_OF_THE_ABOVE(3);
```
angelozerr, line 23: I will rename `NONE_OF_THE_ABOVE` with `NONE `or `UNKWOWN`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
```diff
@@ -0,0 +1,31 @@
...
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.commons;
+
+/**
+ * Represents the context where the cursor is in a Java file.
+ */
+public enum JavaCursorContext {
+
+	EMPTY_FILE(1),
```
angelozerr, line 21: I will rename `EMPTY_FILE` with `ON_EMPTY_FILE`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
```diff
@@ -0,0 +1,31 @@
...
+	CLASS(2),
```
angelozerr, line 22: I will remove this enum and add:

 * `ON_BODY_CLASS` : triggered on a body class, but not on a body method
 * `ON_ANNOTATION_CLASS` : triggered before on a class where annotation can be inserted.
 *  `ON_ANNOTATION_METHOD` : triggered before on a method where annotation can be inserted.
 *  `ON_ANNOTATION_FIELD`:  triggered before on a field where annotation can be inserted.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
```diff
@@ -370,6 +371,42 @@ public Hover hover(MicroProfileJavaHoverParams params, IJDTUtils utils, IProgres
...
+	/**
+	 * Returns the cursor context for the given file and cursor position, or NONE_OF_THE_ABOVE if none of the cursor contexts are applicable.
+	 *
+	 * @param params the completion params that provide the file and cursor position to get the context for
+	 * @param utils the jdt utils
+	 * @param monitor the progress monitor
+	 * @return the cursor context for the given file and cursor position, or null if none of the cursor contexts are applicable
+	 * @throws JavaModelException when the buffer for the file cannot be accessed or the Java model cannot be accessed
+	 */
+	public JavaCursorContext snippetCursorContext(MicroProfileJavaCompletionParams params, IJDTUtils utils,
```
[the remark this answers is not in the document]
angelozerr: Instead of returning the JavaCursorContext (which should be renamed with `JavaCursorContextKind`), I will create a new class JavaCursorContextResult which will host:

 - the kind (ON_CLASS_ANNOTATION, etc)
 - in case of annotation, the offset (or better the position) where the import of annotation type must be done if class doesn't import the annotation type. After that, the snippet completion could return too the additionalTextEdit to insert the proper import.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
```diff
@@ -152,24 +153,29 @@ public CompletableFuture<Either<List<CompletionItem>, CompletionList>> completio
...
-							return ((SnippetContextForJava) context).isMatch(projectInfo);
-						}
-						return true;
-					}, projectInfo) //
-					.forEach(item -> {
-						list1.getItems().add(item);
+
+			CompletableFuture<Void> snippetsFuture = microprofileLanguageServer.getLanguageClient().getJavaSnippetCursorContext(javaParams) //
+					.thenAccept((javaCursorContext) -> {
+						documents.getSnippetRegistry().getCompletionItems(document, finalizedCompletionOffset, canSupportMarkdown,
```
angelozerr, line 164: I wonder if we could avoid using completion offset (finalizedCompletionOffset) and just using the completion position coming from the LSP an dthe completion offset could be resolved on JDT LS side (@jeffmaury is it a problem for you to work with Position (line, char) instead of completion offset on IJ side).

Why I would like to do that? It is because I would like to try to remove the TextDocument for Java file, which maintains offset, although it is just used for completion. We will gain a lot in memory (for big Java file)

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
datho7561: Sorry, I don't understand the `ON_ANNOTATION_CLASS`, `ON_ANNOTATION_METHOD`, `ON_ANNOTATION_FIELD`. If you are in a body class, then aren't you always before a field, a method, or a class? I think that all the annotations that we have snippets for are for methods.

datho7561: ...unless you are in a class with nothing in the body :facepalm:. Okay makes sense.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
datho7561: I don't seem to be having any problems serializing the enum, is this needed?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
datho7561: Let me know what you think of the new naming scheme

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
datho7561: I think this is an interesting idea, and you should file an issue for it.

The important thing to consider is then we also need some mechanism to synchronize the tab and newline information.

angelozerr: Indeed we need those information. If we decide to remove TextDocument, `JavaCursorContextResult` should return those `line information` and if we do that, it should return too the `packageName`.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
angelozerr: Indeed it serialize the enum, but with name. If you provide `forValue` static method, LSP4J serializes the enum by using the int value to have a minimal JSON response size. 

You can see that any LSP4J enum provides this method. See for instance https://github.com/eclipse/lsp4j/blob/5605ca82252833777dbef82ca13a403dbfb9d906/org.eclipse.lsp4j/src/main/java/org/eclipse/lsp4j/DiagnosticSeverity.java#L46

datho7561: The enum is represented as an integer in the JSON-RPC message without adding `forValue`.

datho7561: I see what you mean now, the enum is serialized as a number, but not as the same number

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContext.java
rgrunber: You can also annotate parameters of a method declaration. Is it worth adding that ? I think only Qute has an example of that.

datho7561: Unless we are planning on adding a snippet that supports a specific annotation in the parameters, I don't think it's worthwhile. What is the example in Qute?

rgrunber: https://github.com/redhat-developer/quarkus-ls/issues/637 . Probably not something we'd be looking into for the time being.

datho7561: Yeah, since it's just a single member annotation, I don't think it's worth making a snippet for

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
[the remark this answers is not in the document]
datho7561: I opened https://github.com/eclipse/lsp4mp/issues/339 based on this discussion. Also, here is a branch with some of my initial work that I removed from this PR: https://github.com/datho7561/lsp4mp/tree/import-offset-for-snippets

datho7561:
Alright, the code should be clean and ready for review. Sorry, it took a while to get the code working, since there are a lot of edge cases with nested classes.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContextKind.java
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * The cursor is in a file that does not have a root type declaration.
+	 */
+	IN_EMPTY_FILE(1),
+
+	/**
+	 * The cursor is before a type declaration body, either at the root of a file or
+	 * within another class. The cursor is before any annotations on the type.
+	 */
```
angelozerr, line 29: Could you add please a sample for each enum like

```
|
class A
```

Having a sample is sometimes better than a description. If youhave the both, it is better again :)

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/snippets/SnippetContentType.java
```diff
@@ -0,0 +1,43 @@
...
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.snippets;
+
+/**
+ * Represents the type of the content of a snippet.
+ */
+public enum SnippetContentType {
+	CLASS(1), METHOD(2), FIELD(3), METHOD_ANNOTATION(4);
```
angelozerr, line 20: Move each enum on a new line.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/resources/org/eclipse/lsp4mp/snippets/jakarta-rest.json
```diff
@@ -24,7 +24,8 @@
       "type": [
         "javax.ws.rs.GET",
         "jakarta.ws.rs.GET"
-      ]
+      ],
+      "contentType": "CLASS"
```
angelozerr, line 28: I would prefer writing `"contentType": "class"`.

datho7561: What do you think of using Gson to serialize/deserialize? This information is not sent between microprofile-ls and the jdt component.

angelozerr: If you wish, but I think it should be better to write `"contentType": "class"`

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/snippets/JavaFileCursorContextTest.java
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	@Test
+	public void testEmptyFileContext() throws Exception {
+		IJavaProject javaProject = loadMavenProject(MicroProfileMavenProjectName.config_hover);
+		IProject project = javaProject.getProject();
+		IFile javaFile = project.getFile(new Path("src/main/java/org/acme/config/Empty.java"));
+		String javaFileUri = javaFile.getLocation().toFile().toURI().toString();
+
+		MicroProfileJavaCompletionParams params = new MicroProfileJavaCompletionParams(javaFileUri, new Position(0, 0));
```
angelozerr, line 59: Could you write please as comment for each test where the cursor is triggered. In this sample it should be

// |

microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/snippets/JavaFileCursorContextTest.java
```diff
@@ -0,0 +1,239 @@
...
+	@Test
+	public void testJustSnippetFileContext() throws Exception {
+		IJavaProject javaProject = loadMavenProject(MicroProfileMavenProjectName.config_hover);
+		IProject project = javaProject.getProject();
+		IFile javaFile = project.getFile(new Path("src/main/java/org/acme/config/Empty.java"));
+		String javaFileUri = javaFile.getLocation().toFile().toURI().toString();
+		javaFile.refreshLocal(IResource.DEPTH_ZERO, null);
+		javaFile.setContents(new ByteArrayInputStream("rest_class".getBytes()), 0, MONITOR);
+
+		MicroProfileJavaCompletionParams params = new MicroProfileJavaCompletionParams(javaFileUri, new Position(0, 0));
```
angelozerr, line 73: In this sample it should be:

`|rest_class`

angelozerr: It should be nice to have 2 more tests with:

*  rest_|class
*  rest_class|

angelozerr: Please add this kind of comment for each tests.

angelozerr:
The PR seems working pretty well. To be consistent with Java annotation completion,I think we should show annotation snippet only when completion offset is after an `@` to avoid showing thoses snippets when offset is before the  `@` :

image


 *  `|@` should hide annotation snippets
 * `@|` should show annotation snippets

angelozerr:
I have the impression that this PR breaks the completion method that we provide for fallback method.

I think we should improve the completion participant which should return à completionresult which stores the completionitem list and javacursorcontext both

If we do that we dont need a new delegate command h1ndler for cursor.

datho7561:
> ... I think we should show annotation snippet only when completion offset is after an `@` to avoid showing thoses snippets when offset is before the `@` :

Alright, I support this use case now, by adding a field to the cursor context response that contains what's to the left of the cursor. eg.

```java
public static|
```

I respond with `"static"`

datho7561:
Alright, I also implemented combining the request with the existing `microprofile/java/completion` request. I think this change makes sense, and it makes the code much cleaner. However, with this change we can't get the completion results directly from the java language server component anymore, since then we would have to call the method twice (once in the language server client, and once in microprofile-ls).

angelozerr:
Let me see the code in detail, but my idea is that mp ls consumes completion participant and manage the both usecase

angelozerr:
Great work @datho7561 !

PR #340, opened 2023-01-11
Fix diagnostics data deserialization

### LINKED ISSUE #303: NoValueAssignedToPropertyQuickFix insert property with wrong name
When a defined property is not assigned, this fix allows to insert the property assignment in a property file.
But if this property is part of a group (ConfigProperties) then the quickfix will not use the group prefix but only the property name.

See https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/projects/maven/config-quickstart/src/main/java/org/acme/config/UnassignedValue.java

datho7561:
Diagnostics data is used for some code actions, meaning bad deserialization causes the code actions to break. Added unit tests for the buggy helper method.

Fixes #303


rgrunber (approved):
LGTM. Good catch.

PR #345, opened 2023-01-23
Update eclipse.jdt.ls to 1.20.0-SNAPSHOT

datho7561:

### DIFF
--- a/microprofile.jdt/pom.xml
+++ b/microprofile.jdt/pom.xml
@@ -24,7 +24,7 @@
     <jacoco.destFile>${project.build.directory}/jacoco.exec</jacoco.destFile>
     <sonar.jacoco.reportPath>${jacoco.destFile}</sonar.jacoco.reportPath>
     <surefire.timeout>600</surefire.timeout>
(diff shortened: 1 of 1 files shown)


rgrunber (approved):
There are some infra issues at Eclipse, but the change looks fine to me.

`org.eclipse.jdt.ls.tp @ ${jdt.ls.version}` will get the version of the JDT-LS target platform. This will **NOT** include `org.eclipse.jdt.s.core` . That comes from https://download.eclipse.org/jdtls/snapshots/repository/latest/ . So this just ensures that we use the proper dependencies.

datho7561:
Okay. Do you think it makes sense to wait on the infrastructure to be back up before merging this?

rgrunber:
Sure, we can wait, and monitor on https://www.eclipsestatus.io/incidents/j724xpdbz3sb & https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2549

angelozerr:
Thanks @datho7561 !

PR #351, opened 2023-02-01
Validate `defaultValue` for lists and arrays

datho7561:
- Fix a bug with validating `char` and `java.lang.Character`
- Add a new error message for list-like config properties that have `defaultValue=""`, since the TCK doesn't require this and smallrye-config doesn't support this

Fixes redhat-developer/vscode-microprofile#143

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
@@ -26,6 +26,7 @@ import java.text.MessageFormat;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -71,7 +72,11 @@ public class MicroProfileConfigASTValidator extends JavaASTValidator {
 
 	private static final AntPathMatcher pathMatcher = new AntPathMatcher();
 
+	private static final Pattern ARRAY_SPLITTER = Pattern.compile("(?<!\\\\),");
+
 	private static final String EXPECTED_TYPE_ERROR_MESSAGE = "''{0}'' does not match the expected type of ''{1}''.";
+
+	private static final String EMPTY_LIST_LIKE_WARNING_MESSAGE = "''defaultValue=\"\"'' will behave as if no default value is set, and will not be treated as an empty ''{0}''.";
 
 	private static final String NO_VALUE_ERROR_MESSAGE = "The property ''{0}'' is not assigned a value in any config file, and must be assigned at runtime.";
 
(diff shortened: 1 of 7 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
```diff
@@ -224,36 +225,70 @@ private boolean isPropertyIgnored(String propertyName) {
...
 
+	private static final Pattern ARRAY_SPLITTER = Pattern.compile("(?<!\\\\),");
+
 	private static boolean isAssignable(ITypeBinding fieldBinding, IJavaProject javaProject, String defValue) {
 		String fqn = Signature.getTypeErasure(fieldBinding.getQualifiedName());
+		// handle list-like types.
+		// MicroProfile config supports arrays, `java.util.List`, and `java.util.Set` by
+		// default. See:
+		// https://download.eclipse.org/microprofile/microprofile-config-2.0/microprofile-config-spec-2.0.html#_array_converters
+		if (fieldBinding.isArray()) {
```
rgrunber, line 236: When an `ITypeBinding` is an array, does it work to call `getElementType()` to return the element type of the array ? At least to avoid the string manipulation here.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java
```diff
@@ -224,36 +225,70 @@ private boolean isPropertyIgnored(String propertyName) {
...
+			String itemsTypeFqn = fqn.substring(0, fqn.indexOf('['));
+			for (String listItemValue : ARRAY_SPLITTER.split(defValue, -1)) {
+				listItemValue = listItemValue.replace("\\,", ",");
+				if (!isAssignable(itemsTypeFqn, listItemValue, javaProject)) {
+					return false;
+				}
+			}
+			return true;
+		} else if ("java.util.List".equals(fqn) || "java.util.Set".equals(fqn)) {
+			String itemsTypeFqn = Signature.getTypeErasure(fieldBinding.getTypeArguments()[0].getQualifiedName());
```
rgrunber, line 246: `getElementType()` would simplify this if it works as expected.

JessicaJHee (commented):
Works well! I was just wondering if we also want to allow empty char/int array/lists? Seems like it is not currently allowed. I get:
```
'' does not match the expected type of 'List<Char>'.microprofile-config(DEFAULT_VALUE_IS_WRONG_TYPE)
```

datho7561:
Good catch! That's a bug. I'll fix that first thing tomorrow

rgrunber:
I think this might get tricky.

```
@ConfigProperty(name = "lights", defaultValue = "")
private int [] lights;
```

is not supported by Quarkus itself.

On the other hand,

```
@ConfigProperty(name = "lights", defaultValue = "1 , 2 , 3")
private int [] lights;
```

does seem to work, but we don't support it. I wonder if a simple `StringTokenizer` can be used to support the whitespace separation.

datho7561:
> is not supported by Quarkus itself. If you try to use the variable, you'll get a runtime error.

:exploding_head: I am going to try with `List<Integer>` to see if that works.

> does seem to work, but we don't support it. I wonder if a simple `StringTokenizer` can be used to support the whitespace separation.

Looks like `smallrye-config` trims the leading and trailing whitespace before converting for most number values. I was trying to base the implementation just off the spec, but it doesn't specify what to do in that case.

datho7561:
> is not supported by Quarkus itself. If you try to use the variable, you'll get a runtime error.

I'll add a separate error message for empty array/list/set, because I think this behaviour is unintuitive, and deserves a bit more explanation than "cannot convert".

datho7561:
Actually read the code, and it seems having `defaultValue=""` means don't set a default value. I'll adjust my wording accordingly

rgrunber (approved):
LGTM.

JessicaJHee (approved):
LGTM!

PR #353, opened 2023-02-08
Cancel delegate commands for workspace symbols

datho7561:
When workspace symbol cancel is requested, cancel the delegate commands that compute the workspace symbols on the java language server side.


rgrunber (approved):
Works well for me.

PR #360, opened 2023-02-15
Cancel workspace symbols futures without a new extra Thread.

angelozerr:
Cancel workspace symbols futures without a new extra Thread.

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/CompletableFutureWrapper.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/CompletableFutureWrapper.java
@@ -1,0 +1,349 @@
+/*******************************************************************************
+* Copyright (c) 2023 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.utils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


datho7561:
There is a compilation error, remove this line: https://github.com/eclipse/lsp4mp/pull/360/files#diff-d4c8d8f330417a88c31b0a24c0330d06af21adae8b81620377ac4d10ab8a5a6aR101

I don't think this works. If I open workspace symbols in VS Code then hit `Escape`, I see the workspace symbol cancel notification, but I don't see the cancel notifications for the delegate command requests.

angelozerr:
Ok thanks for your feedback @datho7561 Let me investigate why it doesn't work.

angelozerr:
I debug the JDT workspace symbol (monitor.isCanceled()) and it should work now.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/CompletableFutureWrapper.java
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	@Override
+	public void checkCanceled() {
+		if (delegate.isCancelled())
+			throw new CancellationException();
+	}
+
+	@Override
+	public <T> CompletableFuture<T> cancelIfNeeded(CompletableFuture<T> futureToCancel) {
```
datho7561, line 56: Eclipse gives a warning here, since the type of the future you are adding could be different than the type of the delegate future. It might be better to name the type parameter something other than `T` so it's easier to tell that's what's happening.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/CompletableFutureWrapper.java
```diff
@@ -0,0 +1,294 @@
...
+		this.futuresToCancel.add(futureToCancel);
+		return futureToCancel;
+	}
+
+	/**
+	 * Cancel the delegate future and the list of future to cancel.
+	 */
+	public boolean cancel(boolean mayInterruptIfRunning) {
```
datho7561, line 64: I think it would be nice to add the `@Override` to all the methods. There should be a clean up to do this quickly. If you don't think it's that important, feel free to ignore this comment.

datho7561 (approved):
Works really well! I can see the cancel notifications being sent to jdt.ls now in the language server trace. I added a few comments about some things in the code that I think could be improved, but feel free to merge if you disagree, since the improvement is working.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/utils/CompletableFutureWrapper.java
angelozerr: fixed

datho7561 (approved):

PR #363, opened 2023-02-23
Fix NPE on shutdown when LSP client doesn't define extendedClientCapabilities

angelozerr:
.In [JBossTools Quarkus](https://github.com/jbosstools/jbosstools-quarkus) they don't define the extendedClientCapabilities which cause an NPE when the MP LS is shutdown.

This PR fixes this problem. The code is the same than Qute Language Server.


angelozerr:
@jeffmaury @sbouchet this PR is for you :)

angelozerr:
> @jeffmaury @sbouchet this PR is for you :)

My bad you define it, but this PR takes care of NPE.

datho7561 (approved):
Looks good. I confirmed this works with vscode-microprofile by commenting out the extended client capabilities.

datho7561:
hmm, it looks like something is broken with the build

datho7561:
@angelozerr when you get a change, do you think you could rebase this PR? I think that should fix the tests.

angelozerr:
> @angelozerr when you get a change, do you think you could rebase this PR? I think that should fix the tests.

done

PR #364, opened 2023-02-23
Update to 1.21.0-SNAPSHOT version of JDT-LS target.

rgrunber:


datho7561 (approved):
Looks good! I can confirm the language server still works after this change.

rgrunber:
@datho7561 I was thinking if we should immediately bump LSP4J to 0.20.0 since this is what JDT-LS is using ? https://github.com/eclipse/lsp4mp/blob/6f2e1830b2a5ad12bf7956f8ce392f20db000bb8/microprofile.ls/org.eclipse.lsp4mp.ls/pom.xml#L32 . However 0.14.0 does support the 3.17 spec, and the main overlap is with the delegate commands which should be safe.

datho7561:
I think it'd be nice to upgrade, but not necessary.

PR #365, opened 2023-02-23
Adjust "Unknown property" to "Unrecognized property"

### LINKED ISSUE #290: "Unknown property" maybe should be described as "unused"?
**EDIT:**  I updated this after seeing #289 opened to avoid duplication and only cover what's left of my original issue:

## unused vs. unknown

I think the "unused" adjective might be a better description than "unknown".

My first reaction is: "what do you mean it's unknown?   I'm telling you this is the value."  :)  

OTOH, in Eclipse or VSCode, if I just declare a local variable 'x' without referencing it, I get the message (guessing from JDT LS?): 
`The value of the local variable 'x' is not used.`

So "unused" or to spell it out more "appears to be unused in a MicroProfile Config injection into a Java class" (if it's worth spelling it out more... though not sure it is).

datho7561:
Fixes #290

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileValidator.java
@@ -145,7 +145,7 @@ class PropertiesFileValidator {
 			// The unknown validation must be ignored for this property name
 			return;
 		}
-		addDiagnostic("Unknown property '" + propertyName + "'", property.getKey(), severity,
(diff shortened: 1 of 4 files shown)


datho7561:
@scottkurz what do you think of this rewording?

scottkurz:
> @scottkurz what do you think of this rewording?

I personally still like "unused" better than "unrecognized" because I'm familiar with it from Eclipse and VSCode for e.g. an unused local variable.   But you did elaborate in the addition at the end of the message... so if this is a compromise of all the thoughts around this then it seems like an improvement and it's OK with me.    Thank you for working on this.

rgrunber:
I think the change makes sense so feel free to merge when ready. I noticed that for a property in a Java file that is not defined anywhere we have a different warning so that one is fine.

PR #366, opened 2023-02-23
Error for GraphQL void mutations

### LINKED ISSUE #359: GraphQL: `void` `@Mutation`s are also not allowed
I missed it while reading the specification, but methods annotated with `@Mutation` cannot have a return type of `void`. It would be nice to provide an error for this case.

See https://download.eclipse.org/microprofile/microprofile-graphql-1.0/microprofile-graphql.html#_void_mutations, https://github.com/eclipse/lsp4mp/issues/348

datho7561:
Closes #359


JessicaJHee (approved):
LGTM!

PR #370, opened 2023-03-07
Fix test failures

### LINKED ISSUE #369: Tests error in `BasePropertiesManagerTest#createParentFolders` when running locally
To reproduce the issue, run `mvn clean verify` in `microprofile.jdt`

datho7561:
Use `FORCE` flag when creating folders in the integration tests for `microprofile.jdt`.

Fixes #369


JessicaJHee (approved):
LGTM! Fixes the errors as expected, thank you!

rgrunber:
Makes sense. If the folders already exist there's going to be issues, which is always a risk when generating so many of these test cases. Was the method call throwing an exception ?

datho7561:
Yes, it was throwing an Exception, so the tests would error out

PR #371, opened 2023-03-07
Add an `id` to track the type of each `CodeAction`

datho7561:
Add a new `id` field in the CodeAction data to identify the type of each CodeAction, without having to rely on the display text of the code action.

This is very helpful for telemetry, since it enables collecting info on which CodeActions are the most used.


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.Objects;
+
+/**
+ * Represents the data that all code actions have.
+ */
+public class CodeActionData {
+
+	public static enum CodeActionId {
+		UnknownPropertySimilarTextSuggestion, //
```
angelozerr, line 24: I suggest that CodeActionId stores the title too, il xill simplify the code.

IN other words i suggest  that you see https://github.com/redhat-developer/quarkus-ls/blob/4dc6eb085250d33baf3f0707d1027f3655cf473f/qute.ls/com.redhat.qute.ls/src/main/java/com/redhat/qute/services/diagnostics/QuteErrorCode.java

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,103 @@
...
+		ConfigPropertyInsertDefaultValue, //
+		InsertMissingHealthAnnotation, //
+		InsertInjectAndRestClientAnnotations, //
+		InsertInjectAnnotation, //
+		InsertRegisterRestClient, //
+		InsertRestClientAnnotation, //
+		InsertApplicationScopedAnnotation, //
+		ImplementHealthCheck, //
+		GenerateOpenApiAnnotations, //
+		AssignValueToProperty, //
```
angelozerr, line 38: AssignValueToProperty;

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
```diff
@@ -77,13 +80,22 @@ public CodeAction resolveCodeAction(JavaCodeActionResolveContext context) {
 		ChangeCorrectionProposal proposal = new InsertAnnotationAttributeProposal(name, context.getCompilationUnit(),
 				annotation, 0, attributeName);
 		try {
-			toResolve.setEdit(context.convertToWorkspaceEdit(proposal));
+			WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
```
angelozerr, line 83: Why you need to do that?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationMissingQuickFix.java
```diff
@@ -100,6 +101,7 @@ public CodeAction resolveCodeAction(JavaCodeActionResolveContext context) {
 				context.getASTRoot(), parentType, 0, resolveAnnotationsArray);
 		try {
 			WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
+			we.setChanges(null);
```
angelozerr, line 104: why you need to do that?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/NoValueAssignedToPropertyQuickFix.java
```diff
@@ -134,6 +135,7 @@ public CodeAction resolveCodeAction(JavaCodeActionResolveContext context) {
 		TextDocumentEdit tde = insertTextEdit(new TextDocumentItem(uri, "properties", 0, insertText), insertText,
 				new Position(0, 0));
 		WorkspaceEdit workspaceEdit = new WorkspaceEdit(Collections.singletonList(Either.forLeft(tde)));
+		workspaceEdit.setChanges(null);
```
angelozerr, line 138: why you need to do that?

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,103 @@
...
+package org.eclipse.lsp4mp.commons;
+
+import java.util.Objects;
+
+/**
+ * Represents the data that all code actions have.
+ */
+public class CodeActionData {
+
+	public static enum CodeActionId {
```
angelozerr, line 23: The main problem with enum is that it i snot extensable. As some extension can contribute to code action, the STring id should be used.

We have the similar usecase in lemminx (with diagnostic code). The main idea is that we provide an IXMLErrorCode API and we implement it with enum, see https://github.com/eclipse/lemminx/search?q=IXMLerrorcode

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
datho7561: The reason I did this was so that the two `CodeActionFactory.java` classes are the same. The `microprofile.jdt` one was making `edit` an empty map, and the `microprofile.ls` one was making `edit` `null`. I think it makes sense to be consistent about which one CodeActions have, so I chose to set it to `null` everywhere. Unfortunately, the jdt.ls utils all set `edit` to an empty map, so I remove it here.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
datho7561: I started doing this, but not every CodeAction has an associated template. For instance, there are many different code actions that insert an annotation (insert health annotation, insert inject annotation, ...) but there is only one template used for all these different code actions, and it's used in an abstract class. Since there isn't a 1 to 1 mapping of code action type to title template, I'm not sure this is worthwhile.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
datho7561: Okay, I've implemented an extensible id system

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,76 @@
...
+		return id;
+	}
+
+	/**
+	 * Sets the id of this code action.
+	 *
+	 * @param id the new value for the id of this code action
+	 */
+	public void setCodeActionId(MicroProfileCodeActionId id) {
+		if (id != null) {
```
datho7561, line 53: I should clear it if it's null

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Represents the data that all code actions have.
+ */
+public class CodeActionData {
+
+	private String id;
+
+	public CodeActionData() {
+		this.id = null;
```
angelozerr, line 26: `this(null);`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	public CodeActionData(ICodeActionId id) {
+		if (id != null) {
```
angelozerr, line 30: replace with `setCodeActionId(id);`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/CodeActionData.java
```diff
@@ -0,0 +1,78 @@
...
+	public String getCodeActionId() {
+		return id;
+	}
+
+	/**
+	 * Sets the id of this code action.
+	 *
+	 * @param id the new value for the id of this code action
+	 */
+	public void setCodeActionId(MicroProfileCodeActionId id) {
```
angelozerr, line 52: `public void setCodeActionId(ICodeActionId id) {`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
angelozerr: What is the problem if changes is an empty map?

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/commons/ICodeActionId.java
```diff
@@ -0,0 +1,28 @@
...
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+* which is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+*
+* Contributors:
+*     Red Hat Inc. - initial API and implementation
+*******************************************************************************/
+package org.eclipse.lsp4mp.commons;
```
angelozerr, line 14: I will move all code action classes in `org.eclipse.lsp4mp.commons.codeactions`

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/java/codeaction/InsertAnnotationAttributeQuickFix.java
datho7561: There is no difference in behaviour between an empty map and a null entry, but the tests will fail if there is an empty map and a null entry is expected or vice versa.

datho7561:
Okay, sorry about the delay, I undid the change where I was setting the workspace edit changes to null. I use a helper in the tests to set it to null if it's empty.

angelozerr:
LGTM. Thanks @datho7561 !

PR #377, opened 2023-03-23
Fix definition for property value of user-defined class

### LINKED ISSUE #374: `textDocument/definition` fails on property value
To reproduce:

create a file that defines a property `my.property`:

```java
@ConfigProperty(name="my.property")
Integer myProperty;
```

Then, in `microprofile-config.properties`, add the following:

```properties
my.property=1234
```

Then, invoke definition on `1234`. The request should fail. The expected result is that nothing happens, since it makes no sense to get the definition of  a value.

datho7561:
Prevent an Exception when finding the definition of a property value by checking if the field that the value refers to actually exists in the compilation unit before trying to get its source range.

Fixes #374


datho7561:
@JessicaJHee if you have time, do you think you could review this PR?

JessicaJHee (approved):
LGTM! Solves the issues as expected

PR #332, opened 2022-12-09
Use template to determine jakarta namespace in snippets

datho7561:
- Use `${jakarta}` to point to either `javax` or `jakarta`, depending on which is available, and prefering `jakarta`
- Rename `jaxrc` to `rest_class` and `jaxrm` to `rest_get`

Closes #320, closes #229

Signed-off-by: David Thompson <davthomp@redhat.com>

PR #378, opened 2023-03-23
Changelog for 0.7.0

datho7561:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # MicroProfile Language Server Changelog
 
+## [0.7.0](https://github.com/eclipse/lsp4mp/milestone/7?closed=1) (April 3, 2023)
+
+### Enhancements
+ * Validation for GraphQL `@Query` and `@Mutation` methods with `void` return type. See [#348](https://github.com/eclipse/lsp4mp/issues/348), [#359](https://github.com/eclipse/lsp4mp/issues/359).
+ * Navigate to REST endpoints using workspace symbols. See [#87](https://github.com/eclipse/lsp4mp/issues/87).
+ * Show config property documentation when hovering over the key in a properties file. See [#321](https://github.com/eclipse/lsp4mp/issues/321).
+ * Validate lists in `@ConfigProperty`'s `defaultValue`. See [#351](https://github.com/eclipse/lsp4mp/pull/351).
+ * Make Java file snippets context-aware. See [#108](https://github.com/eclipse/lsp4mp/issues/108).
+ * Add an `id` to track the type of each `CodeAction`. See [#371](https://github.com/eclipse/lsp4mp/pull/371).
+
+### Bug Fixes
+ * Hover fails in properties files when the Java language server is loading. See [#375](https://github.com/eclipse/lsp4mp/issues/375).
+ * Definition sometimes fails on property values in a properties file. See [#374](https://github.com/eclipse/lsp4mp/issues/374).
+ * Adjust go to definition range for property keys to include the offset between the property key and `=`. See [#335](https://github.com/eclipse/lsp4mp/pull/335).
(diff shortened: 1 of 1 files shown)


angelozerr:
LGTM

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # MicroProfile Language Server Changelog
 
+## [0.7.0](https://github.com/eclipse/lsp4mp/milestone/7?closed=1) (March 28, 2023)
+
+### Enhancements
+ * Validation for GraphQL `@Query` and `@Mutation` methods with `void` return type. See [#348](https://github.com/eclipse/lsp4mp/issues/348), [#359](https://github.com/eclipse/lsp4mp/issues/359).
+ * Navigate to REST endpoints using WorkspaceSymbols. See [#87](https://github.com/eclipse/lsp4mp/issues/87).
```
rgrunber, line 7: I would maybe call it "workspace symbols". Doing that means if we refer to this feature on the extension side, it's a mere copy/paste, as opposed to having to rename it bcause "WorkspaceSymbols" is not really a term we should exposing to users.

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+ * Show config property documentation when hovering over the key in a properties file. See [#321](https://github.com/eclipse/lsp4mp/issues/321).
+ * Validate lists in `@ConfigProperty`'s `defaultValue`. See [#351](https://github.com/eclipse/lsp4mp/pull/351).
+
+### Bug Fixes
+ * Hover fails in properties files when the Java language server is loading. See [#375](https://github.com/eclipse/lsp4mp/issues/375).
+ * Definition sometimes fails on property values in a properties file. See [#374](https://github.com/eclipse/lsp4mp/issues/374).
+ * Only show Java file snippets relevant to the context. See [#108](https://github.com/eclipse/lsp4mp/issues/108).
```
rgrunber, line 14: What are your thoughts on making this a feature ? It's pretty nice. Could reword it to "Make Java file snippets context aware" .

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+ * Adjust go to definition range for property keys to include the offset between the property key and `=`. See [#335](https://github.com/eclipse/lsp4mp/pull/335).
+ * Fix NullPointerException during go to definition in properties files. See [#372](https://github.com/eclipse/lsp4mp/issues/372).
+ * Fix NullPointerException on shutdown when LSP client doesn't define extendedClientCapabilities. See [#363](https://github.com/eclipse/lsp4mp/pull/363).
```
rgrunber, line 17: Could surround NPE in back-ticks, or just shorte to NPE.

Also, if something is a NPE, it's nicer to state the fix in terms of any user behaviour that wasn't possible with the NPE and is possible once fixed. If the NPE didn't really break things in a meaningful way (other than it polluted the logs), then we can keep it as-is.

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+ * Completion causes Exceptions when typing in a Java file. See [#347](https://github.com/eclipse/lsp4mp/issues/347).
+ * Support projects that use classes from the `jakarta` namespace (JakartaEE 9). See [#344](https://github.com/eclipse/lsp4mp/issues/344).
```
rgrunber, line 19: What about https://github.com/eclipse/lsp4mp/pull/332, which closes : https://github.com/eclipse/lsp4mp/issues/229, https://github.com/eclipse/lsp4mp/issues/320 ? If they should be mentioned, does it make sense to turn this entry into some general "Support the Jakarta namespace" ?

rgrunber (commented):
Also what about https://github.com/eclipse/lsp4mp/pull/353 , https://github.com/eclipse/lsp4mp/pull/371 ?

datho7561:
#353 is a bug fix for a feature introduced in this release. I guess I could include #371 under enhancements, but it by itself is not a feature.

PR #381, opened 2023-04-11
Fix context-aware snippets in projects that use Project Lombok

datho7561:
Fixes redhat-developer/vscode-microprofile#155


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
```diff
@@ -794,4 +802,20 @@ private static int offsetOfFirstNonAnnotationModifier(BodyDeclaration node) {
...
+	 * otherwise.
+	 * 
+	 * @param node the node to check if it's generated
+	 * @return true if the given node was generated by Project Lombok and false
+	 *         otherwise
+	 */
+	private static boolean isGenerated(ASTNode node) {
+		try {
+			return ((Boolean) node.getClass().getField("$isGenerated").get(node)).booleanValue();
+		} catch (IllegalArgumentException | IllegalAccessException | NoSuchFieldException | SecurityException e) {
```
JessicaJHee, line 816: Do we have to catch `IllegalArgumentException` and `SecurityException` here?

JessicaJHee (approved):
Looks great and solves the issue when tested! Just had one small question.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManagerForJava.java
datho7561: Yes, it looks like it's required in order to access fields reflexively.

Issue #389, opened 2023-04-27
Improve completion performance in properties files


JessicaJHee:
This requires two changes:
1. Implement resolve for documentation
2. Implement `itemDefaults` for `completionList` to reduce the size of the completion result. See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionList

angelozerr:
I suggest that you start to implement the 1.

JessicaJHee:
From my testing, the performance improvement in the LS response time is as follows:

|  | 1st run | 2nd run | 3rd run | Average|
| ------------- | ------------- | ------------- | ------------ | ------------ | 
| Before change  | 137ms | 141ms | 118ms | 132ms |
| After change | 91ms | 85ms | 88ms | 88ms |

Outside of this, there is a visible difference on the client side to actually display the completion result since the `textDocument/completion` response is much smaller now.

angelozerr:
Thanks @JessicaJHee for youŕ feedback. Great job!

PR #392, opened 2023-05-02
Provide completionItem/resolve support

JessicaJHee:
Part of fix for #389

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/properties/PropertiesFileTextDocumentService.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/properties/PropertiesFileTextDocumentService.java
@@ -55,6 +55,7 @@ import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4mp.commons.MicroProfileProjectInfo;
 import org.eclipse.lsp4mp.commons.MicroProfileProjectInfoParams;
 import org.eclipse.lsp4mp.commons.MicroProfilePropertiesChangeEvent;
+import org.eclipse.lsp4mp.commons.utils.JSONUtility;
 import org.eclipse.lsp4mp.ls.AbstractTextDocumentService;
 import org.eclipse.lsp4mp.ls.MicroProfileLanguageServer;
 import org.eclipse.lsp4mp.ls.api.MicroProfileLanguageServerAPI.JsonSchemaForProjectInfo;
@@ -62,6 +63,7 @@ import org.eclipse.lsp4mp.ls.commons.ModelTextDocument;
 import org.eclipse.lsp4mp.ls.commons.ModelTextDocuments;
 import org.eclipse.lsp4mp.ls.commons.ValidatorDelayer;
 import org.eclipse.lsp4mp.model.PropertiesModel;
+import org.eclipse.lsp4mp.services.properties.CompletionData;
 import org.eclipse.lsp4mp.services.properties.PropertiesFileLanguageService;
 import org.eclipse.lsp4mp.settings.MicroProfileFormattingSettings;
 import org.eclipse.lsp4mp.settings.MicroProfileInlayHintSettings;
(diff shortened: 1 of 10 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/properties/PropertiesFileTextDocumentService.java
```diff
@@ -456,4 +479,39 @@ public <R> CompletableFuture<R> getPropertiesModelCompose(TextDocumentIdentifier
...
+	 * @param data      the data entry field comings from the CompletionItem,
+	 *                  CodeAction, etc.
+	 *
+	 * @param fieldName the entry field name (ex : "uri", "propertyName").
+	 *
+	 * @return the property value of the given <code>fieldName</code> if the given
+	 *         <code>data</code> and null otherwise.
+	 */
+	public static String getProperty(Object data, String fieldName) {
+		JsonObject json = JSONUtility.toModel(data, JsonObject.class);
```
datho7561, line 509: Instead of doing this, you can make a Java class called `CompletionData` with fields `propertyName` and `uri` and a no-argument constructor.

Then you can call `JSONUtility.toModel(data, CompletionData.class)`, and it will return an instance of `CompletionData`.

There is an example of this at the end of `InitializationOptionsSettings.java`.

I think this will make it a bit easier to read the code.

datho7561 (changes requested):
Works very well, a very noticable performance improvement! I have a comment about how the completion resolve data is serialized/deserialized, but once that's addressed, it should be good to merge

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/properties/PropertiesFileTextDocumentService.java
angelozerr: I agree with @datho7561 it is better to do that since completion support is not extensible with participant like lemminx.

angelozerr:
@JessicaJHee your PR looks very good. I wonder if it is possible with LSP to define your data as default and not for each completion items. If it is possible it would be nice to use it this strategy only if lsp client have this default data capability

JessicaJHee:
> @JessicaJHee your PR looks very good. I wonder if it is possible with LSP to define your data as default and not for each completion items. If it is possible it would be nice to use it this strategy only if lsp client have this default data capability

I believe I can do this with `itemDefaults`. I was planning to do this in another PR that implements other fields (like `editRange` and `insertTextFormat` ) in the `itemDefaults` as well if that sounds good?

angelozerr:
> I believe I can do this with itemDefaults. I was planning to do this in another PR that implements other fields (like editRange and insertTextFormat ) in the itemDefaults as well if that sounds good?

yes sure!

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/CompletionData.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	/**
+	 * Returns the minimal data to resolve a completion item.
+	 *
+	 * @param propertyName the property name of the completion item.
+	 * @param uri          the uri of the text document.
+	 * @return the minimal data to resolve a completion item.
+	 */
+	public static JsonObject createCompletionData(String propertyName, String uri) {
```
datho7561, line 63: You should be able to set `data` directly as an instance of `CompletionData`. It will be automatically turned into JSON when the message is serialized.

Here's my suggestion:
- add a constructor for `CompletionData` that accepts `propertyName` and `uri`, and use this instead of `createCompletionData(...)`
- add a no-argument constructor that sets `propertyName` and `uri` to `null` (this is needed for the JSON serialization to work)

datho7561 (changes requested):

datho7561 (approved):
Looks good. Thanks, Jessica!

PR #394, opened 2023-05-03
Issue 390: Refactor javaCursorContext calls

### LINKED ISSUE #390: make javaCursorContext calls consistent with all PropertiesManagerForJava calls
in 0.7.1, there is a new javaCursorContext method in PropertiesManagerForJava that can be called statically.
as there is a singleton pattern involved in this class, that may break the contract to directly call this method.
please remove static and rezfactor clients to use getInstance() calls.

sbouchet:
fixes #390


angelozerr:
It misess changes with getInstance() in some tests like https://github.com/eclipse/lsp4mp/blob/8952fdf1eb17a470251b32fe8b76d209be80b7cb/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/snippets/JavaFileCursorContextTest.java#L64

sbouchet:
last commit looks better ;)

angelozerr:
Thanks @sbouchet !

PR #395, opened 2023-05-05
Add completionList itemDefaults support in properties files

### LINKED ISSUE #389: Improve completion performance in properties files

JessicaJHee:
Fixes #389

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
@@ -27,6 +27,7 @@ import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.CompletionItemDefaults;
 import org.eclipse.lsp4j.CompletionItemKind;
 import org.eclipse.lsp4j.CompletionList;
 import org.eclipse.lsp4j.InsertTextFormat;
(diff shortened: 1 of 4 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/PropertiesFileCompletions.java
```diff
@@ -596,4 +630,29 @@ protected String getExpr(TextDocument document, int completionOffset) {
...
+		return completionItemDefaults;
+	}
+
+	private void initializeCompletionListItemDefaults(PropertiesModel model,
+			MicroProfileCompletionCapabilities completionCapabilities, int offset) {
+		CompletionItemDefaults itemDefaults = getCompletionItemDefaults();
+		boolean snippetsSupported = completionCapabilities.isCompletionSnippetsSupported();
+		if (completionCapabilities.isCompletionListItemDefaultsInsertTextFormatSupport()) {
+			itemDefaults.setInsertTextFormat(snippetsSupported ? InsertTextFormat.Snippet : InsertTextFormat.PlainText);
+		}
```
datho7561, line 646: Put these steps in `getCompletionItemDefaults()` so that you only do them once. I realize you will need to set the range every time you respond to a completion request, but once the server is initialized, the `isCompletionSnippetsSupported()` shouldn't change.

datho7561 (changes requested):
One small change that I think would be nice, but otherwise works great!

datho7561 (approved):
Alright, it looks good, I will merge it! Thanks, Jessica!

PR #411, opened 2023-07-13
Remove unnecessary fields in properties file completion result

### LINKED ISSUE #410: Improve again completion result of microprofile-config.properties
Here a sample of completion item result:

```json
{
      "label": "quarkus.management.filter.{*}.matches",
      "kind": 10,
      "filterText": "quarkus.management.filter.${1:key}.matches",
      "textEditText": "quarkus.management.filter.${1:key}.matches=$0",
      "data": {
          "propertyName": "quarkus.management.filter.{*}.matches",
          "uri": "file:///c%3A/Users/azerr/git/quarkus-blast/src/main/resources/application.properties"
      }
  }
```

I think we can improve again the result to have a lower JSON result since completion for microprofile-config.properties can return a lot of items.

It should be nice to return only:

```json
{
      "label": "quarkus.management.filter.{*}.matches",
      "kind": 10,
      "textEditText": "quarkus.management.filter.${1:key}.matches=$0"
  }
```

* filterText can be removed and LSP cient will use label instead which will be better since filterText contains snippet syntax ("filterText": "quarkus.management.filter.${1:key}.matches"
* data/propertyName is the same value than label, so we can remove it and use label.
* data/uri are the same for all completion items, so we can move this information in CompletionItemDefaults/data

@JessicaJHee what do you think about that?

JessicaJHee:
Fixes #410

* This PR removes `filterText` and `data.propertyName` where `label` is used instead
* `CompletionData.uri` is left since vscode client (not sure about intellij or other clients) does not support `itemDefaults.data` indicated by client capabilities.

These are the supported fields for `itemDefaults` in vscode:
```
                "completionList": {
                    "itemDefaults": [
                        "commitCharacters",
                        "editRange",
                        "insertTextFormat",
                        "insertTextMode"
                    ]
                }
```

Here is an example of the completion item result after the change: 
```
        {
            "label": "quarkus.cache.enabled",
            "kind": 10,
            "textEditText": "quarkus.cache.enabled=${1|false,true|}",
            "data": {
                "uri": "file:///home/jhe/Projects/test_files/quarkus-quickstarts/cache-quickstart/src/main/resources/application.properties"
            }
        },
```

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/CompletionData.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/services/properties/CompletionData.java
@@ -34,8 +34,7 @@ public class CompletionData {
 		this.uri = null;
 	}
 
-	public CompletionData(String propertyName, String uri) {
-		this.propertyName = propertyName;
+	public CompletionData(String uri) {
 		this.uri = uri;
 	}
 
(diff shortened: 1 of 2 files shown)


angelozerr:
It works like a charm @JessicaJHee , good job!

PR #415, opened 2023-08-01
Update eclipse.jdt.ls to 1.27.0-SNAPSHOT

rgrunber:
- Automatic JVM detection in JDT Debug seems to have introduced a regression in JDT indexing causing java.lang.OutOfMemory, so use DetectVMInstallationsJob.disabled=true property from JDT Debug

I noticed the OOM issue when running locally. If it isn't happening on the Eclipse builds, it's likely because the JDK version that is causing it, is not present.


datho7561 (approved):
Looks good. Thanks, Roland!

PR #420, opened 2023-08-31
Fix canProvideJaxRsMethodInfoForClass for default JAX-RS

angelozerr:
Fix canProvideJaxRsMethodInfoForClass for default JAX-RS


datho7561 (approved):
Good catch! Thanks, Angelo!

PR #423, opened 2023-09-15
Use Maven 3.9.4 in the MicroProfile language server.

rgrunber:
From https://github.com/eclipse/lsp4mp/pull/422, I missed also doing the same for `microprofile.ls/org.eclipse.lsp4mp.ls`.


fbricon:
Maven 3.9.4 is out https://maven.apache.org/docs/3.9.4/release-notes.html
Maybe use the [distributionType](https://maven.apache.org/wrapper/maven-wrapper-plugin/wrapper-mojo.html#distributiontype) flag to generate the java source downloader

PR #432, opened 2024-01-26
fix: use newer JDT.LS's JavadocContentAccess2 methods

fbricon:
Fixes https://github.com/redhat-developer/vscode-quarkus/issues/638
But Qute-LS will need a similar fix.

Signed-off-by: Fred Bricon <fbricon@gmail.com>


rgrunber (approved):
This makes lsp4mp build again so just based on that, it's the right thing to do. `JavadocContentAccess2` should be accessible also in jdt.core.manpulation via. newer JDT-LS target platform.

PR #433, opened 2024-02-01
fix: don't generate a symbol name when property starts with '.'

angelozerr:
fix: don't generate a symbol name when property starts with '.'

Fixes https://github.com/redhat-developer/vscode-quarkus/issues/640


fbricon:
@angelozerr thanks!

PR #435, opened 2024-02-01
Temporarily disable tests for jdt component in order to release

datho7561:
We would like to release. The jdt component tests are flakey, and these test failures are preventing the release build from completing.


rgrunber (approved):
This might as well be the norm for now given they are unstable. As long as we can confirm they work locally, or as part of the pr verification I'm fine with skipping in the release.

PR #444, opened 2024-05-15
fix: synchronized projectLabels + collect codeLens when project is loaded.

angelozerr:
fix: synchronized projectLabels + collect codeLens when project is loaded.


fbricon:
thanks @angelozerr!

PR #447, opened 2024-06-13
Update jdt.ls

### LINKED ISSUE #446: Update to the latest jdt-ls
I think we need to use the latest jdt.ls, since the i builds that the older version references have been deleted.

datho7561:
Fixes #446


rgrunber (approved):

datho7561:
I don't know if this new failure is legit or not

datho7561:
I think it might actually be broken; let me try using it

datho7561:
It seems to be working in my IDE, rerunning the tests to make sure!

PR #460, opened 2024-09-09
feat: Provide property validator extension API

angelozerr:
feat: Provide property validator extension API


datho7561 (approved):
Looks good. Thanks, Angelo!

PR #462, opened 2024-09-20
fix : prevent from empty name symbol

angelozerr:
fix : prevent from empty name symbol

Given this microprofile-config.properties file which have 2 dots in property:

`a..b=`

vscode crashes because symbol name must be never empty.

This PR prevent from non empty name symbol.


angelozerr:
@datho7561 if you have time, could you review this PR please.

It is very annoying because if you write a..b= you have popup error.

datho7561 (approved):
Looks good and prevents the error. Thanks, Angelo!

datho7561:
Trying to see if the test failure is legitimate.

PR #464, opened 2024-10-09
fix: split register completion options for Java / Properties

angelozerr:
fix: split register completion options for Java / Properties

This PR allows to register options with given languageId 

1. to avoid sending LSP request in Java file:

 * for InlayHint
 * for documentSymbol
 * for formatting / rangeFormatting (already done)
 * for documentHighlight (already done)

2. customize completion trigger characters and resolveCompletion between microprofile-config.properties and Java files

For instance when you write `{` in Java file it should not trigger completion (today you can see that it triggers completion in Java file when you type `{`

It is not annoying for vscode but very annoying for IJ (see issue at https://github.com/redhat-developer/intellij-quarkus/issues/1395)

As we did some languageId filter with `microprofile-properties` languageId , we need to add in the filter documentSelector the `quarkus-properties` languageId to avoid loosing features like inlay hint, dpcument symbol. The PR https://github.com/redhat-developer/quarkus-ls/pull/999 covers that.

There is a bug in application.properties which doesnt manage highlight when you click on properties. I have fixed this bug too in https://github.com/redhat-developer/quarkus-ls/pull/999

### DIFF
--- a/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/MicroProfileLanguageIds.java
+++ b/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/MicroProfileLanguageIds.java
@@ -1,0 +1,27 @@
+/*******************************************************************************
+* Copyright (c) 2024 Red Hat Inc. and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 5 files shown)


angelozerr:
@datho7561 do you think you could have time to review my PR and the same for Quarkus https://github.com/redhat-developer/quarkus-ls/pull/999 ?

The main thing which is important that we need to check that I have not broken some feature like inlayhint, codelens, completion etc

datho7561 (approved):
Looks good to me, I tested a few different features and they still work.

angelozerr:
> Looks good to me, I tested a few different features and they still work.

Thanks a lot!

PR #467, opened 2024-10-15
REST endpoint codelens is not resolving a `public static final` parameter

angelozerr:
REST endpoint codelens is not resolving a `public static final` parameter

See https://github.com/redhat-developer/intellij-quarkus/issues/1393

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java
@@ -19,12 +19,14 @@ import java.util.regex.Pattern;
 import org.eclipse.jdt.core.IAnnotatable;
 import org.eclipse.jdt.core.IAnnotation;
 import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IField;
 import org.eclipse.jdt.core.IImportContainer;
 import org.eclipse.jdt.core.IImportDeclaration;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IMemberValuePair;
 import org.eclipse.jdt.core.ISourceRange;
 import org.eclipse.jdt.core.ISourceReference;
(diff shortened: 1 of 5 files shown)


angelozerr:
@rgrunber it seems my PR is working. I need to write tests.

If you have time please review the JDT code if you think it is correct. Thanks!

angelozerr:
@rgrunber it seems my PR is working. I need to write tests.

If you have time please review the JDT code if you think it is correct. Thanks!

angelozerr:
@rgrunber as all tests are OK and as I need this PR for Quarkus LS, I merge the PR.

If  you see some corner uscase or some problem @rgrunber please tell me. Thanks.

rgrunber:
A minor bug that can be fixed later. : Something like `import static org.acme.rest.client.Extension.EXTERNAL_PATH;` where  `EXTERNAL_PATH` will be referenced as a simple name (eg. `@Path(EXTERNAL_PATH)`) but `getFieldValue(..)` will return `null`. I guess you might want to try the `K_QUALIFIED_NAME` "logic" after `getFieldValue(..)` if it returns `null`.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/utils/AnnotationUtils.java
```diff
@@ -281,7 +283,68 @@ public static boolean isMatchAnnotation(Annotation annotation, String annotation
...
+					case IMemberValuePair.K_QUALIFIED_NAME: {
+						// ex : class SomePage {
+						// private static final String PATH = "foo";
+						// @Path(SomePage.PATH) --> returns foo
+						IType type = ((IType) annotation.getAncestor(IJavaElement.TYPE));
+						if (type != null) {							
+							int index = memberValue.lastIndexOf(".");
+							String fieldName = memberValue.substring(index + 1);
+							String valueClass = memberValue.substring(0, index);
+							String[][] resolvedTypes = type.resolveType(valueClass);
```
rgrunber, line 313: I guess the only other question I would have had is whether it's necessary to call `resolveType` here. Can't you pass `valueClass` directly to `findType(..)` from below ? It should be a fully qualified type after all.

PR #470, opened 2024-10-17
Improve the Target platform for JDT Extension components.

rgrunber:
We can merge this after @angelozerr does the release. We would need to update the platform I-build repo just like JDT-LS does, unless we automate it, or avoid using the I-build location here.

- Use the JDT-LS repository, which now contains all dependencies to build the target platform
- Include o.e.platform.feature.group for the test runtime
- The target file can now be used both for building and development

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/snippets/JavaFileCursorContextTest.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/snippets/JavaFileCursorContextTest.java
@@ -16,10 +16,13 @@ package org.eclipse.lsp4mp.jdt.core.snippets;
 import static org.junit.Assert.assertEquals;
 
 import java.io.ByteArrayInputStream;
+import java.io.File;
 
+import org.apache.commons.io.FileUtils;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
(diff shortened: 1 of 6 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.test/src/main/java/org/eclipse/lsp4mp/jdt/core/snippets/JavaFileCursorContextTest.java
```diff
@@ -42,11 +46,10 @@ public class JavaFileCursorContextTest extends BasePropertiesManagerTest {
 
 	@After
 	public void cleanUp() throws Exception {
```
rgrunber, line 48: I actually think this "cleanup" needs to be moved into `BasePropertiesManagerTest`. If you have 2 or more testcases that import the exact same sample project, the sample project is copied into place. The previous `${project-dir}/.settings/` folder disappears. It  gets generated only by the project import operation. However, the `.settings/` folder doesn't get regnerated because the Eclipse project model still contains the stale project. The import doesn't occur and things like project compiler settings are incorrect (ie. `JavaCore.COMPILER_COMPLIANCE` is `23` (global default) instead of `1.8` (pom configured)).

The solution is to delete the project not just on the filesystem, but in the Eclipse project model as well.

However, looking around, I noticed **many** other projects that don't do this cleanup, and they didn't seem to have any issues with compiler options. Actually, something as simple as `project.refreshLocal(IResource.INFINITE, null)` seems to trigger a refresh which exposes the problem (compiler options disappear). So we really need to future-proof this.

rgrunber: I've decided to address this in a separate contribution. I know roughly which 2 or 3 tests are failing but I'd like to address it in a separate commit.

datho7561 (approved):

datho7561:
I no longer see the 'missing java classes' failures in the tests after this change, so I think it's a good idea to merge this ASAP.

PR #471, opened 2024-10-21
Update test project compliances to at least 1.8

datho7561:
Eclipse no longer supports Java 7 and earlier. By reimporting the test projects, Eclipse has auto updated their compliances to be at least 1.8 .


datho7561:
This doesn't fix the issue as I thought it would

PR #473, opened 2024-10-22
Configure version of tycho-versions-plugin to use

datho7561:
Use 4.0.8; there appears to be [a bug](https://github.com/eclipse-tycho/tycho/issues/4381) in 4.0.9 related to missing classes


rgrunber:
I basically just did https://github.com/eclipse/lsp4mp/commit/d2bc141f7fb024d668e8788b0572748ad7caca00 as a quick workaround.

datho7561:
Okay, I thought you did that as a temporary change in the Jenkins interface. If it's already committed, I don't think we need this PR.

PR #476, opened 2024-11-04
Do not allow snippets of type method or field before a class statement

### LINKED ISSUE #402: Invalid snippet suggested in completion before the Class statement [context]
This issue revolves around the field and method snippets and the context in which the language server suggests them when the user activates completion (e.g. Ctrl-Space).

Consider this sample class:
```
package a.b.c;
// Position A
class d{ 
    // Position B
    class Inner {}
    // Position C
}
```
When you use completion in Position A the language server provides `rest_get` as one of the choices. This snippet represents the insertion of a method but this is not valid in this context. When you use completion in Position C the snippet `rest_get` is valid. It looks like method snippets should not be allowed before a class. This requires fix #1.

This brings us to Position B. Currently the language server considers this to be "before a class." We must allow a method snippet here. This exposes an ambiguity in the definitions. Position B is both 'In A Class' and 'Before A Class.' The former should dominate the latter and this would allow the language server to suggest a method snippet here. This requires fix #2.

turkeylurkey:
The issue is that `rest_get` and `mpirc` snippets are provided as suggestions when the user's cursor is between the `package` statement and the `class` statement. This is incorrect and it is easy for a person to stumble across.

This fix helps with this problem but it means these two snippets do not appear when the cursor is just before an `enum` or an inner class statement. This seems like an improvement. 

Is this change acceptable?

Fixes #402

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContextKind.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/JavaCursorContextKind.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2023 Red Hat Inc. and others.
+* Copyright (c) 2023, 2025 Red Hat Inc. and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
@@ -29,20 +29,8 @@ public enum JavaCursorContextKind {
 	IN_EMPTY_FILE(1),
 
 	/**
-	 * The cursor is before a type declaration body, either at the root of a file or
-	 * within another class. The cursor is before any annotations on the type.
-	 *
-	 * eg.
-	 * <code>
-	 * <br />
-	 * package org.acme;<br />
-	 * <br />
-	 * |<br />
-	 * &commat;Inject<br />
-	 * public class MyClass {<br />
-	 * <br />
-	 * }<br />
-	 * </code>
+	 * The cursor is before a type declaration body within another class.
+	 * The cursor is before any annotations on the type.
 	 *
 	 * eg.
 	 * <code>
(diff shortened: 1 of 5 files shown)


angelozerr:
I understand it is an annoying problem but we need to see if in JDT with vscode microprofile we have the same problem.

@datho7561 what do you think about this change?

angelozerr:
I will test too in IJ Quarkus if we have the same problem and if we can do something on IJ side.

datho7561:
This issue with this change is this case:

```
package com.github.datho7561;

/**
 * My documentation
 */
public class MyObject {

	public int getValue() {
		return 12;
	}

	|

	static public class MyInnerClass {

	}
}
```

The snippet `mpirc` (for creating a rest client) should be available here, but it isn't. Having a descriptor `BEFORE_CLASS` is useful for annotations that can only be applied to classes (and interfaces and records and such). However, in order to properly handle completion, we need to distinguish between `BEFORE_INNER_CLASS` and `BEFORE_TOP_LEVEL_CLASS`. Currently, in the JDT code, there are two locations where we return `BEFORE_CLASS`; one of them should be changed to `BEFORE_TOP_LEVEL_CLASS` and the other to `BEFORE_INNER_CLASS`. Then we can use this more specific information to better inform snippet completion results.

angelozerr:
Thanks  @datho7561 for your relevant information. IJ need to improve your usecase.

turkeylurkey:
I removed the previous suggestion since it was only a partial solution. I integrated the information from @datho7561 above and implemented a proper solution using the constants `BEFORE_INNER_CLASS` and `BEFORE_TOP_LEVEL_CLASS`.

I tested this language server jar using my IntelliJ plugin and also the `vscode-microprofile` plugin. The field and method snippets are no longer offered between the package statement and the class statement.

@angelozerr  and @datho7561 , could you please review this fix and let me know?

Next we will need a release of lsp4mp and vscode-microprofile. Thank you for your attention.

angelozerr:
@datho7561 is there any chance that you could review the PR please?

datho7561:
Oops just testing the tests to make sure they are passing (or that they aren't failing any worse than before). Didn't realize CI wasn't running.

datho7561:
I ran the tests locally and none of the failures seem related to this PR, so I'll merge.

turkeylurkey:
Thanks, folks.

PR #479, opened 2024-12-05
Use progress monitor when finding types in project

datho7561:
This allows for nested types to be resolved. Without a progress monitor, JDT doesn't resolve nested types.


rgrunber (approved):
I think it's worth doing this. I don't see any tests fixed by this in microprofile.jdt but we know at least 2 instances in quarkus-ls that would benefit.

PR #480, opened 2024-12-05
Potential fix for `JDTTypeUtils.findType()`

datho7561:
https://github.com/redhat-developer/quarkus-ls/commit/f8561effd6c12024bbfb5f8eaa4b90c0aeb4967d caused some regressions in the quarkus-ls tests, since some types were no longer found by JDTTypeUtils.findType(). This attempts to fix it again.

Try the 'secondary types' search first, and if that fails, fall back to the old search we were doing.


datho7561:
I've tested against quarkus-ls locally, and it appears to be working there. However, I wasn't able to reproduce the initial bug from the test run locally, so we won't know for sure until we build on CI.

PR #481, opened 2024-12-06
Improve test performance and stability.

rgrunber:
- For projects with multiple testcases that are importing the same project, simply perform the import once in @BeforeClass and re-use
- Remove .classpath / .project from microprofile-applicationpath test project
- Ensure the Java AST Validator contributions are loaded earlier


datho7561 (approved):
Seems like a great improvement. There's 2 tests that are still flaky and 1 that's consistently failing (the Javadoc one), but the tests are running much faster on Jenkins.

PR #485, opened 2025-02-20
Use Java 21 in build

### LINKED ISSUE #484: Build requires Java 21
jdt.core manipulation has a BREE of Java 21, so we need to use Java 21 in the build. Currently, Java 11 is configured, so the build fails.

datho7561:
Fixes #484

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
   agent any
   tools {
-    jdk 'temurin-jdk17-latest'
+    jdk 'temurin-jdk21-latest'
   }
   environment {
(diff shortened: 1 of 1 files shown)


datho7561:
I used Jenkins' replay feature to run the build with this Jenkinsfile here: https://ci.eclipse.org/lsp4mp/job/lsp4mp/job/master/363/

It seems to not stop at the error about the BREE. I'll wait for that build to finish to see if anything else is wrong, but otherwise, I'll merge this PR.

datho7561:
There are two failing tests, but this gets it past the BREE failure

PR #486, opened 2025-02-20
Enable required extended capabilities when loading project

### LINKED ISSUE #450: Tests failing
Some tests are failing, preventing a snapshot from being released

datho7561:
Fixes #450

TODO per-merge:
- [ ] Remove GitHub Action

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,36 @@
+
+name: Continuous Integration Testing
+on:
+  push:
+    branches:
+      - main
+  pull_request: {}
+jobs:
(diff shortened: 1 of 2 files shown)


datho7561:
I threw in a GitHub Action to test that the changes work. I expect at least 1 failure, since the link in the Bulkhead docs is hardcoded to the Jenkins maven repo path

datho7561:
It seems the diagnostic tests are still flakey.

I think we need to make a decision if we are okay with testing the Bulkhead link knowing it will only work properly on Jenkins, or if we want to disable the extended capability for that test run and pretend the link doesn't exist.

PR #487, opened 2025-02-24
Do not test for link in Bulkhead property

datho7561:
The link is different on every computer. There is no trivial way to only check for the link's presence. With the current test setup, jdt-ls will not resolve the link, since we aren't enabling the extended capability properly.

Fixes a part of #450


datho7561:
Expecting ~3 tests failing since they are flakey (usually related to validation)

datho7561:
I think I figured out how to fix the rest of the flaky tests... just need to test to make sure it's actually working...

datho7561:
@rgrunber  if you have a chance, could you please review this PR? This should fix one of the failing tests

rgrunber:
Fixes the testcase locally for me.

PR #492, opened 2025-04-10
Guard against annotation binding failing to resolve.

rgrunber:
Wasn't able to reproduce a case of this, but I have seen stacktraces that clearly show :

```
SEVERE: Error while visiting node with org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceASTValidator
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.core.dom.IAnnotationBinding.getAnnotationType()" because "binding" is null
    at org.eclipse.lsp4mp.jdt.core.utils.AnnotationUtils.isMatchAnnotation(AnnotationUtils.java:268)
    at org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceASTValidator.validateMethod(MicroProfileFaultToleranceASTValidator.java:179)   
    at org.eclipse.lsp4mp.jdt.internal.faulttolerance.java.MicroProfileFaultToleranceASTValidator.visit(MicroProfileFaultToleranceASTValidator.java:123)
    at org.eclipse.lsp4mp.jdt.internal.core.java.validators.MultiASTVisitor.visit(MultiASTVisitor.java:118)
```


datho7561 (approved):
Wouldn't know how to reproduce, but looking at the source code it's possible that `resolveAnnotationBinding` returns `null`, so it's good to guard against.

rgrunber:
I think it's just taking too long and timing out.

PR #494, opened 2025-04-14
Skip making a fake project if only sources are being scanned

datho7561:
This change provides a modest performance improvement, since the fake Java project is only needed to scan the dependencies.

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManager.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManager.java
@@ -316,6 +316,9 @@ public class PropertiesManager {
 	 */
 	private IJavaProject getJavaProject(IJavaProject javaProject, boolean excludeTestCode,
 			List<MicroProfilePropertiesScope> scopes, SubMonitor monitor) throws JavaModelException {
+		if (scopes.indexOf(MicroProfilePropertiesScope.dependencies) == -1) {
+			return javaProject;
+		}
 		if (javaProject instanceof FakeJavaProject) {
 			// The java project is already resolved
 			return javaProject;


datho7561:
I have no idea if this is correct or if this will actually make the language server more efficient.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/PropertiesManager.java
```diff
@@ -320,6 +320,9 @@ private IJavaProject getJavaProject(IJavaProject javaProject, boolean excludeTes
 			// The java project is already resolved
 			return javaProject;
 		}
+		if (scopes.indexOf(MicroProfilePropertiesScope.dependencies) == -1) {
```
angelozerr, line 323: I will do this test at the begin of configureSearchClasspath method.

datho7561:
> It is the idea that I had in my mind.

Okay, that's a good sign. I've tested adding config properties in Java files; when I do that I can access those properties in completion. I also tried removing a config property definition from the Java file while it was set in the application.properties, and on save the warning appeared in the application.properties.

angelozerr:
> > It is the idea that I had in my mind.
> 
> Okay, that's a good sign. I've tested adding config properties in Java files; when I do that I can access those properties in completion. I also tried removing a config property definition from the Java file while it was set in the application.properties, and on save the warning appeared in the application.properties.

That it is great! Could you just check that the project that you receive is not a FakeJavaProject?

If yes, do you feel some performance improvement?

datho7561:
The IJavaProjects I'm getting when I hit the breakpoint on save are all real

I think I feel a bit of a performance improvement. It feels like it takes diagnostics updating from a quarter second to a bit over half a second. (when I have ~10 source files open in a ~500 source file project).

I'll get some real timing data though; I can calculate the time between `didSave` and the last `publishDiagnostics`...

datho7561:
In both cases, based on the client/server trace, it's around 1 second (but the timestamps only go down to 1 second of resolution). I'd need to dig deeper to figure out the real time difference.

angelozerr:
Anyway it is a good fix because it avoids to twe1k classpath when only java files changed.

PR #496, opened 2025-04-17
CHANGELOG for 0.14.0

datho7561:
We have several bug fixes and a hopefully meaningful performance improvement that are unreleased, and we haven't released in a few months, so it would be nice to do a release. I set the planned release date to Tuesday, as several of us have a 4 day weekend that would not be good to release right before.


rgrunber:
The performance improvement alone would probably make it worthy of a minor release.

PR #498, opened 2025-04-22
Upversion to 0.14.2

datho7561:


datho7561:
I'm checking out the test failures; they are almost always test timeouts on Jenkins. That's worrying, because the test timeout is set to 10 minutes _for each test case_, whereas locally on my computer the entire suite consistently finishes in just over 10 minutes. Something is happening that's specific to the Jenkins environment.

datho7561:
I think the root cause of the tests timing out is that the agents do not have sufficient memory.

I'm going to merge this PR, since this shouldn't affect test failures and is needed to proceed with new features and bug fixes.

PR #499, opened 2025-04-22
Add GitHub Action -based CI

### LINKED ISSUE #495: Seperate reporting for each test suite
In quarkus-ls (which shares a similar project structure, since it extends lsp4mp), I made a change recently that reports the results from the different test suites separately. This is helpful, because the tests for the jdt component are flaky, which means we often don't check (or don't even run!) the tests for the language server component.

In quarkus-ls I accomplished with with parallel GitHub jobs. Currently we use the Eclipse Jenkins for PR CI in this repo, but we could consider only using Jenkins for releasing and using GitHub Actions to run the PR CI.

It might also be possible to do this "multiple reporting" with Jenkins, I think there are Jenkins plugins to add separate entries into the GitHub CI reporting, and we can configure the build to always run all the test suites even if the previous ones failed. This might be a good route, considering the tests tend to pass more consistently on Jenkins, likely due to the agents having more resources (CPU, memory)

datho7561:
There are two goals for this PR:
- I think that the root cause of the test timeout on Jenkins is that not enough RAM is available to the Jenkins agents
- We can run the microprofile.jdt and microprofile.ls test suites in parallel and report their results separately
  - In the existing Jenkins suite, if the microprofile.jdt test suite fails, we don't run the microprofile.ls suite, which is bad because the microprofile.jdt almost always fails, so we never know the results of the microprofile.ls suite

Fixes #495

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,0 +1,29 @@
+name: Tests
+on: [push, pull_request]
+jobs:
+  microprofile-jdt-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+      - name: Set up Eclipse Temurin JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+      - name: Build and test lsp4mp JDT component
+        working-directory: microprofile.jdt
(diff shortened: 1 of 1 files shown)


datho7561:
I could also look into https://github.com/marketplace/actions/publish-test-results to improve reporting further; let me know if that looks helpful

PR #500, opened 2025-05-14
Don't re-trigger twice Java validation when a Java file is saved.

angelozerr:
Don't re-trigger twice Java validation when a Java file is saved.

This PR avoid sending (type=3 which means configFile changed)

```
[Trace - 9:33:16 PM] Sending notification 'microprofile/propertiesChanged'.
Params: {
    "type": [
        3
    ],
    "projectURIs": [
        "C:\\Users\\azerr\\Downloads\\demo\\demo"
    ]
}
```

which retriggers validation of Java files when a Java file is saved.

See https://github.com/redhat-developer/vscode-quarkus/issues/1004#issuecomment-2881352209

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
@@ -20,6 +20,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceChangeEvent;
 import org.eclipse.core.resources.IResourceChangeListener;
@@ -27,12 +28,15 @@ import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.IResourceDeltaVisitor;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.jdt.core.ElementChangedEvent;
 import org.eclipse.jdt.core.IElementChangedListener;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaElementDelta;
 import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
 import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.lsp4mp.commons.MicroProfilePropertiesChangeEvent;
 import org.eclipse.lsp4mp.commons.MicroProfilePropertiesScope;
 import org.eclipse.lsp4mp.jdt.core.IMicroProfilePropertiesChangedListener;
@@ -146,11 +150,17 @@ public class MicroProfilePropertiesListenerManager {
 			if (resource == null) {
 				return false;
 			}
+
+			// Step 5: Process folders, files, and projects as needed
 			switch (resource.getType()) {
 			case IResource.ROOT:
+				return true;
 			case IResource.PROJECT:
+				IProject project = (IProject) resource;
+				return project.isAccessible() && project.hasNature(JavaCore.NATURE_ID);
 			case IResource.FOLDER:
-				return resource.isAccessible();
+				// Ignore folder which belongs to Java output file location (ex: target/classes)
+				return resource.isAccessible() && !isInOutput(resource);
 			case IResource.FILE:
 				IFile file = (IFile) resource;
 				if (isJavaFile(file) && isFileContentChanged(delta)) {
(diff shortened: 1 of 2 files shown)


microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
```diff
@@ -125,8 +125,17 @@ public void didClose(DidCloseTextDocumentParams params) {
...
+		// which can negatively impact performance.
+		// As there's no comment explaining the reason for this behavior,
+		// we are disabling this feature.
+
 		// validate all opened java files which belong to a MicroProfile project
-		triggerValidationForAll(null);
+		// triggerValidationForAll(null);
+		
+		// Validate only the opened Java file
+		validate(documents.onDidSaveTextDocument(params), true);
```
angelozerr, line 138: @rgrunber it was here that trigger validation for all opened Java files was done. I disable that because I don't see some usecase which requires that. I prefer loosing a not commons usecase (and which usecase?) to have better performance.

angelozerr: And I wonder if we should not trigger validation on didSave 

I dont see any value to retrigger validation on didSave 

@datho7561 @rgrunber what do you think about that?

datho7561: Make sense to me

datho7561 (approved):
This works great and the code changes look good. I'd like to hear from Roland as well, but I think it's a good idea to merge this.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
```diff
@@ -125,8 +125,17 @@ public void didClose(DidCloseTextDocumentParams params) {
...
+		
+		// Since the beginning of the MP LS project,
+		// saving a Java file triggers validation for all open Java files,
+		// which can negatively impact performance.
+		// As there's no comment explaining the reason for this behavior,
+		// we are disabling this feature.
+
 		// validate all opened java files which belong to a MicroProfile project
-		triggerValidationForAll(null);
+		// triggerValidationForAll(null);
```
rgrunber, line 137: I think this breaks the following behaviour. You have a Java source file that defines configuration properties used in the configuration file. If those are added/removed, it affects some diagnostics :

https://github.com/user-attachments/assets/2a21b852-6686-465b-8a79-9f2addccff27

The diagnostics will not update when saving the config file, but they do update when continuing to type in the source file, or re-opening it. Are we ok with that ?

Are we ok with that ?

rgrunber (commented):
Definitely fixes the properties changed notifications. Just noticed one behaviour change we should be aware of. Is this ok ?

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
angelozerr: I remember that I had the similar problem with IJ Quarkus.

I am pretty sure that if you save your application.properties a second time, it will work.

If it that it is because we load in cache properties (for checking validation on Java file) from the /target/application.properties and not from src/main/resources/application.properties. 

I think we should load cache from src/main/resources/application.properties. 

We will loose the computed variable from maven for instance declared in teh properties file, but I have never seen this usecase.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
```diff
@@ -171,6 +181,43 @@ public boolean visit(IResourceDelta delta) throws CoreException {
...
+		 * @throws JavaModelException
+		 */
+		private static boolean isInOutput(IResource resource) throws JavaModelException {
+
+			IJavaProject javaProject = JavaCore.create(resource.getProject());
+			IPath resourcePath = resource.getFullPath();
+
+			// Exclude resources in the main output location (e.g.,/ProjectName/bin)
+			IPath outputLocation = javaProject.getOutputLocation();
+			IPath outputFullPath = ResourcesPlugin.getWorkspace().getRoot().getFolder(outputLocation).getFullPath();
```
rgrunber, line 200: This isn't the full path. 

You could compute the path relative to the workspace, in which case I don't think you need `outputFullPath`. you can do `outputLocation.isPrefixOf(resourcePath)`.

If you want the absolute path on the filesystem, why not do something like `javaProject.getProject().getLocation().append(outputLocation)` which should be the absolute path. I've seen it used in some other places.

microprofile.ls/org.eclipse.lsp4mp.ls/src/main/java/org/eclipse/lsp4mp/ls/java/JavaFileTextDocumentService.java
rgrunber: Saving a second time will not fix this. Validation isn't triggered. You either need to type, or re-open the file.

angelozerr: That's strange it is working for me?

rgrunber: Are you making changes and then saving or just hitting "Save" on an unchanged file ? Making changes will update the diagnostics but saving doesn't. Where would the validation come from when `triggerValidationForAll` is removed from `didSave` above.

rgrunber: Typing a space in the stale Java source file even without the save will update the diagnostics. Look at the video "With this change" above! You can see just typing updates the diagnostics so that's fine.

angelozerr: Sorry I mean tyep a space and save application.properried and do that twice.

It should update java diagnostics correctly, right?

rgrunber: You're right, it does work. Interesting.

angelozerr:
I close this PR in favor of https://github.com/eclipse-lsp4mp/lsp4mp/pull/546

PR #502, opened 2025-07-30
Addresses fix #520: JavaCursorContextKind IllegalArgumentException in lsp4mp

archana-1924:
The PR addresses Issue: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520
Root causes were observed in JavaCursorContextKind.java files in

* lsp4jakarta.ls
* lsp4jakarta.jdt


turkeylurkey:
@archana-1924 I agree, also some code in liberty tools treats this ls and the jakarta ls the same so we likely need to make this change in jakarta.

angelozerr:
Thanks @archana-1924 !

PR #505, opened 2025-09-19
Update target platform to use 4.38 ibuild

datho7561:


datho7561:
The tests for microprofile.ls are passing flawlessly locally, so I'm very confused why they're failing here.

rgrunber (approved):
I think this should be fine since platform itself tends to be more stable, but just be aware that in https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3526 I opted to use the 4.27 release until at least M1 of 4.38 for stability.

Issue #396, opened 2023-05-10
@ConfigProperty defaultValue marked wrongly as an error ...'does not match expected type of '... despite fact that correct org.eclipse.microprofile.config.spi.Converter exists


mariankassovic:
Validation incorrectly reports:
'00:00' does not match the expected type of 'LocalTime'. 

despite fact that there is defined org.eclipse.microprofile.config.spi.Converter for LocalTime

  @Inject
  @ConfigProperty(name = "startTime", defaultValue = "00:00")
  private LocalTime startTime;

converter is implemented as simple as

import java.time.LocalTime;

import org.eclipse.microprofile.config.spi.Converter;

public class LocalTimePropertyConverter implements Converter<LocalTime> {

  @Override
  public LocalTime convert(String value) {
    return LocalTime.parse(value);
  }
}

code works as expected, compiles and junits works as well - only validation reports incorrect error.

angelozerr:
We should have the same problem in JDT. If I remember we skip converter because it is not an easy task to support them and we take care of simple type without converter.

I dont know if it will be easy to support this issue.

E14:
I think we have a misunderstanding - Currently when the validator encounters an unsupported type (a type that lsp4mp does not support validating), like Duration, it will result in `DEFAULT_VALUE_IS_WRONG_TYPE`, which is incorrect.

I would say the culprit is in [org.eclipse.lsp4mp.jdt.internal.config.java.MicroProfileConfigASTValidator#isAssignable(java.lang.String, java.lang.String, IJavaProject)](https://github.com/eclipse-lsp4mp/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/config/java/MicroProfileConfigASTValidator.java#L311)

Quickest fix should be to return `true` on the default case instead, which would effectively "skip" the checks and still report actual failures like before. But I can see value in reporting (as Info or Notice?) when validation is not supported, so users who are used to this validation are notified in some form when that does not happen.

E14:
@angelozerr would you accept PRs for this? I'm happy to contribute, just not if it's going to be rejected

angelozerr:
@E14 yes please create a PR and add 2 screenshots with before/after your fix. Thanks!

E14:
> [@E14](https://github.com/E14) yes please create a PR and add 2 screenshots with before/after your fix. Thanks!

[...]

PR #510, opened 2025-10-22
Extend MicroProfileConfig default-value validation

E14:
Validating a default value can succeed, fail, or be unsupported.

Partly resolves / mitigates #396 

Before:
image

After:
image


E14:
> Could you sign please eca.

Done!

> I think you should adjust some tests or add new tests.

Fully agree - I mostly created this PR to use the CI because master is not stable (😑). Give me a bit though, I took too long today getting a spammable mailbox that both eclipse and github accept so I can connect the ECA

Glad the basic idea looks good to you though!

angelozerr:
@E14 please see the issue https://github.com/eclipse-lsp4mp/lsp4mp/issues/511 I am working on it

PR #513, opened 2025-11-16
fix: Optional indexed property

angelozerr:
fix: Optional indexed property

This PR allows to write optional key or index in a property which have Map or List.

For instance `quarkus.log.category.{*}.level` defines a Map of level. With is PR you can define the property without key (ex: `quarkus.log.category."foo".level`) like:

`quarkus.log.category.level=ERROR`


angelozerr:
@datho7561 I merge this PR in order to consume it on IJ Quarkus plugin and have feedback from user ASAP.

I have written tests.

PR #520, opened 2025-11-24
Use diagnostics from the context

angelozerr:
This PR fixes a design error when diagnostics must be collected. The IJavaDiagnosticsParticipant defines a method like:

`List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProgressMonitor monitor) throws CoreException;`

which returns a list of diagnostics for the given context (a given file uri)

The collectDiagnostics should not return a list of diagnostics and should use `JavaDiagnosticsContext #addDiagnostic`

This PR fixes that and fixes also the file uri which is used when diagnostic is created although diagnostic doesn't store some uri.


angelozerr:
@datho7561 please review this PR. I know it will break the current API but I have not seen some code instead in LSP4MP which implements this API.

@turkeylurkey and your team please tell me if you are OK with this change. I have done this similar changes in IJ Quarkus.

turkeylurkey:
This change is ok for us too.

PR #524, opened 2025-12-04
Handle Javadoc for both old and new jdt.ls versions

datho7561:
Use either `JavadocContentAccess2.getMarkdownContent` or `JavadocContentAccess2.getMarkdownContentReader` depending on what's available.

This is important, since the jdt.ls with the API change is used during compilation and CI, but is not published yet.

To test: build the jdt component and then run it in VS Code using vscode-microprofile.


fbricon:
We can also restore the methods in jdtls

datho7561:
If it simplifies the code, I think it's worth keeping the change in jdt.ls. We're accessing non-API code from jdt.ls here anyways. lsp4mp should only use this messy implementation for a few releases of jdt.ls/vscode-java, then we can switch over to just using the new method.

FYI Angelo was hoping to get a release today.

datho7561:
I opened https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3634; I don't really have a strong opinion on which solution is better, but we will need one of them in order to release today.

fbricon:
Easier to restore the code in jdtls, which I merged, once jdtls is released, you can then migrate to the new code, no need for ugly reflection hacks

PR #546, opened 2026-04-21
Don't re-trigger Java validation for all opened Java files when a Java file is saved.

angelozerr:
Don't re-trigger Java validation for all opened Java files when a Java file is saved.

This PR avoid sending (type=3 which means configFile changed)

```
[Trace - 9:33:16 PM] Sending notification 'microprofile/propertiesChanged'.
Params: {
    "type": [
        3
    ],
    "projectURIs": [
        "C:\\Users\\azerr\\Downloads\\demo\\demo"
    ]
}
```

which retriggers validation of Java files when a Java file is saved.

See https://github.com/redhat-developer/vscode-quarkus/issues/1004#issuecomment-2881352209

# Before this PR

 * when Java file was saved,  'microprofile/propertiesChanged' was sent because `target/classes/application.properties` was updated
 * when src/main/resources/application.properties,  'microprofile/propertiesChanged' was sent twice (one update from `src/main/resources/application.properties` and one update from `target/classes/application.properties`

# After this PR

 * when Java file is saved,  no 'microprofile/propertiesChanged' should be sent.
 * when src/main/resources/application.properties is saved, only one  'microprofile/propertiesChanged' should be sent

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProjectManager.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProjectManager.java
@@ -23,10 +23,7 @@ import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceChangeEvent;
 import org.eclipse.core.resources.IResourceChangeListener;
-import org.eclipse.core.resources.IResourceDelta;
-import org.eclipse.core.resources.IResourceDeltaVisitor;
 import org.eclipse.core.resources.ResourcesPlugin;
-import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
@@ -52,7 +49,10 @@ public class JDTMicroProfileProjectManager {
 	private final Map<IJavaProject, JDTMicroProfileProject> projects;
 	private MicroProfileProjectListener microprofileProjectListener;
 
-	private class MicroProfileProjectListener implements IResourceChangeListener, IResourceDeltaVisitor {
+	/**
+	 * Resource Listener to update MicroProfile projects cache.
+	 */
+	private class MicroProfileProjectListener implements IResourceChangeListener {
 
 		@Override
 		public void resourceChanged(IResourceChangeEvent event) {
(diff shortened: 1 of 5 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/core/project/JDTMicroProfileProjectManager.java
```diff
@@ -104,44 +92,6 @@ private void evict(IProject project) {
...
-				return resource.isAccessible();
-			case IResource.FILE:
-				IFile file = (IFile) resource;
-				if ((isFileDeleted(delta) || isFileContentChanged(delta) || isFileAdded(delta))
-						&& isConfigSource(file)) {
-					// it's a config source file (ex : microprofile-config.properties)
-					JDTMicroProfileProject mpProject = getJDTMicroProfileProject(file);
-					if (mpProject != null) {
-						// Evict the properties cache
-						mpProject.evictConfigSourcesCache();
```
angelozerr, line 126: This code was super annoying because it evicted the config source cache as soon as an application.properties was updated (in other words when a Java file was saved because target/classes/application.properties was updated) or when src/main/resources/application.properties was updated.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
```diff
@@ -189,12 +189,33 @@ public boolean visit(IResourceDelta delta) throws CoreException {
...
+					JDTMicroProfileProject mpProject = JDTMicroProfileProjectManager.getInstance()
+							.getJDTMicroProfileProject(file);
+					
+					boolean generateEvent = false;
+					if (isFileDeleted(delta) || isFileAdded(delta)) {
+						generateEvent = true;
+						// Create, delete config source file (ex : microprofile-config.properties)						
+						if (mpProject != null) {
+							// Evict the properties cache
+							mpProject.evictConfigSourcesCache();
```
angelozerr, line 205: The evict cache of config source is done now only when a config file is created/deleted which is required to rebuild the config files scope.

microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/MicroProfilePropertiesListenerManager.java
```diff
@@ -189,12 +189,33 @@ public boolean visit(IResourceDelta delta) throws CoreException {
...
+						}
+					} else if (isFileContentChanged(delta)) {
+						// Update config source file (ex : microprofile-config.properties)
+						generateEvent = mpProject != null ? mpProject.updateConfigSource(file) : true;
```
angelozerr, line 209: This case comes from when:

 * a Java file is saved, target/application.properties is updated.
 * src/main/resources/application.properties is updated.

The mpProject.updateConfigSource(file) returns true only when src/main/resources/application.properties is updated

When application.properties is updated (only for sources), we don't evict the full cache of config sources but just reset the IConfigSource

datho7561 (approved):
Looks good and makes sense. Thanks, Angelo!

PR #550, opened 2026-07-10
Accept the relaxed `Duration` format (e.g. `60S`) for SmallRye/Quarkus projects in SAFE mode

Eng-Fouad:
## Summary

In SAFE execution mode, validating a `.properties` value whose type is `java.time.Duration` rejects the relaxed duration format used by Quarkus/SmallRye. For example, a `Duration` property such as:

```properties
quarkus.http.read-timeout=60S
```

is flagged with:

> Text cannot be parsed to a Duration

even though it is valid at runtime. This PR makes SAFE mode accept the same relaxed format that Quarkus/SmallRye accept, while keeping strict ISO-8601 validation for runtimes that do not use SmallRye Config (e.g. Open Liberty).

## Root cause

SAFE mode builds an embedded config via `new SmallRyeConfigBuilder().build()` and validates values through the converter returned by `Config#getConverter(Duration.class)`.

SmallRye Config core does **not** register an explicit converter for `java.time.Duration`, so it falls back to the *implicit* converter, which simply calls `java.time.Duration.parse(value)`. That only accepts strict ISO-8601 (`PT60S`), so relaxed values like `60S`, `10m`, or a bare number (`600`, interpreted as seconds) throw `DateTimeParseException`.

The relaxed parsing lives in the Quarkus `DurationConverter`, which is on the project classpath and is therefore only used in FULL mode. SAFE mode never sees it.

## Changes

- Add `DurationConverter` (in `…commons.runtime.converter.safe`) that mirrors the Quarkus/SmallRye rules:
  - a bare number is interpreted as a number of seconds (`600` → `PT10M`);
  - a value starting with a digit is parsed as an ISO-8601 duration without the leading `PT` (`60S` → `PT60S`, `10m` → `PT10M`);
  - any other value is parsed as strict ISO-8601 (`PT1H30M`).
- Register it in `SafeConverterRuntimeSupport#loadConfig()` **only when the project uses SmallRye Config**, detected via a `smallrye-config` entry on the project classpath. This keeps strict ISO-8601 validation for non-SmallRye runtimes (e.g. Open Liberty).
- Add `MicroProfileProjectRuntime#hasClasspathEntry(String)` — inspects classpath entry *names* only (no class is loaded), so it is safe to use in SAFE mode.

FULL mode is unchanged (it already uses the project's own converter).

The change is applied to both copies of the runtime/converter code: `microprofile.ls` (`org.eclipse.lsp4mp.ls`) and `microprofile.jdt` (`org.eclipse.lsp4mp.jdt.core`).

## Why detect `smallrye-config` rather than "Quarkus"

The relaxed duration format is a SmallRye Config behavior, and Quarkus projects always bundle `smallrye-config-*.jar`. The test `QUARKUS_PROJECT_RUNTIME` classpath itself contains `smallrye-config-*.jar` (and no literal `quarkus-*.jar`), so `smallrye-config` is both the semantically correct and the reliably testable signal.

## Testing

New `testDuration` cases assert the behavior in both directions, and the full SAFE/FULL suites still pass (66 tests, 0 failures):

- `QuarkusProjectRuntimeInSafeModeTest#testDuration` — `60S`, `10m`, `3h`, `600`, `PT1H30M` are accepted; `foo` is still reported.
- `LibertyProjectRuntimeInSafeModeTest#testDuration` — `PT1H30M` is accepted, but `60S` is still reported (no SmallRye Config on the classpath).

| Runtime (SAFE mode) | `60S` | `PT1H30M` |
|---|---|---|
| Quarkus (has SmallRye Config) | accepted | accepted |
| Open Liberty (no SmallRye Config) | rejected (`Text cannot be parsed to a Duration`) | accepted |

## Notes

- No public API is removed; `MicroProfileProjectRuntime#hasClasspathEntry(String)` is additive.
- `DurationConverter` is source/target Java 8 compatible (consistent with the module's `maven.compiler.source/target`).

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/runtime/MicroProfileProjectRuntime.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/commons/runtime/MicroProfileProjectRuntime.java
@@ -171,6 +171,30 @@ public class MicroProfileProjectRuntime implements TypeProvider {
 	}
 
 	/**
+	 * Returns {@code true} if the project classpath contains an entry whose name
+	 * contains the given {@code token} (typically an artifact file name).
+	 *
+	 * <p>
+	 * Only the classpath entry names are inspected; no class is loaded, so this can
+	 * safely be used in SAFE execution mode.
+	 * </p>
+	 *
+	 * @param token the substring to look for in the classpath entry names
+	 * @return {@code true} if a classpath entry contains the given token
+	 */
+	public boolean hasClasspathEntry(String token) {
+		if (classpath == null || token == null) {
+			return false;
+		}
+		for (String entry : classpath) {
+			if (entry != null && entry.contains(token)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	/**
 	 * Retrieves the Class corresponding to the given fully qualified name. Uses an
 	 * internal cache to avoid reloading the same class multiple times.
 	 *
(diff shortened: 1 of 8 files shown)


angelozerr:
@Eng-Fouad I think it is annoying to manage Quarkus usecase in MicroProfile LSP44MP. Why you cannot use full mode?

Issue #549, opened 2026-07-08
Upgrade JDT LS to 1.60.0 and lsp4j to 1.0.0


venmanyarun:
All milestone releases on https://download.eclipse.org/jdtls/milestones/ were accidentally deleted due to a category.xml referencing a bundle that Eclipse recently removed. This broke the JDT LS promotion job which deleted all milestone directories. JDT LS 1.60.0 has since been restored and is the new minimum version we need to adopt. See https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3823.

Objective

Upgrade JDT LS from the snapshot you are using to 1.60.0, and upgrade lsp4j to 1.0.0 in lsp4mp

Note that, JDT LS 1.60 requires Java 21

angelozerr:
@venmanyarun any contribution are welcome!

datho7561:
Seems this is addressed, so I'll close this issue.

PR #551, opened 2026-07-30
Upgrade JDT LS to 1.60.0 and lsp4j to 1.0.0

MeherSru:
Updates lsp4mp to use JDT LS 1.60.0 and lsp4j 1.0.0, as requested in #549. This also means Java 21 is now required (JDT LS 1.60 needs it).
Some code had to be updated because lsp4j 1.0.0 changed a few of its APIs.

### DIFF
--- a/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/ChangeUtil.java
+++ b/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/ChangeUtil.java
@@ -112,7 +112,11 @@ public class ChangeUtil {
 			}
 
 			VersionedTextDocumentIdentifier identifier = new VersionedTextDocumentIdentifier(uri, 0);
-			TextDocumentEdit documentEdit = new TextDocumentEdit(identifier, converter.convert());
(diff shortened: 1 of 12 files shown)


microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/core/java/ChangeUtil.java
```diff
@@ -112,7 +112,11 @@ private static void convertTextEdit(WorkspaceEdit root, ICompilationUnit unit, T
 			}
 
 			VersionedTextDocumentIdentifier identifier = new VersionedTextDocumentIdentifier(uri, 0);
-			TextDocumentEdit documentEdit = new TextDocumentEdit(identifier, converter.convert());
+			List<Either<org.eclipse.lsp4j.TextEdit, org.eclipse.lsp4j.SnippetTextEdit>> edits = new ArrayList<>();
```
datho7561: Use `converter.convertToTextDocumentEdit`

datho7561 (changes requested):
Found one spot where you can avoid duplicating code, otherwise looking pretty good

datho7561 (approved):
Looks good! There are some codelens related failures, but I think they're unrelated to this change, and this gets the project building again.

